PR #38, opened 2024-02-10
feat: add protected routing tm-15

### LINKED ISSUE #15: Protected routing
- [x] Create ProtectedRoute wrapper component
- [x] Redirect not signed in users from protected screens to Sign In page

tarmiel:
Resolves #15

### DIFF
--- a/apps/frontend/src/index.tsx
+++ b/apps/frontend/src/index.tsx
@@ -4,6 +4,7 @@ import { createRoot } from "react-dom/client";
 import "~/assets/css/styles.css";
 import {
 	App,
+	ProtectedRoute,
 	RouterProvider,
 	StoreProvider,
 } from "~/libs/components/components.js";
@@ -19,7 +20,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 					{
 						children: [
 							{
-								element: "Root",
+								element: <ProtectedRoute component={"Root"} />,
 								path: AppRoute.ROOT,
 							},
 							{
--- a/apps/frontend/src/index.tsx
+++ b/apps/frontend/src/index.tsx
@@ -20,7 +20,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 					{
 						children: [
 							{
-								element: <ProtectedRoute component={"Root"} />,
+								element: <ProtectedRoute component="Root" />,
 								path: AppRoute.ROOT,
 							},
 							{
--- a/apps/frontend/src/libs/components/protected-route/protected-route.tsx
+++ b/apps/frontend/src/libs/components/protected-route/protected-route.tsx
@@ -1,5 +1,4 @@
-import { Navigate } from "react-router-dom";
-
+import { Navigate } from "~/libs/components/components.js";
 import { AppRoute } from "~/libs/enums/enums.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 import { type ValueOf } from "~/libs/types/types.js";
--- a/apps/frontend/src/libs/components/protected-route/protected-route.tsx
+++ b/apps/frontend/src/libs/components/protected-route/protected-route.tsx
@@ -1,17 +1,15 @@
-import type { FC, ReactNode } from "react";
-
 import { Navigate } from "react-router-dom";
 
-import { AppRoute } from "~/libs/enums/app-route.enum.js";
+import { AppRoute } from "~/libs/enums/enums.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 import { type ValueOf } from "~/libs/types/types.js";
 
 type Properties = {
-	component: ReactNode;
+	component: React.ReactNode;
 	redirectTo?: ValueOf<typeof AppRoute>;
 };
 
-const ProtectedRoute: FC<Properties> = ({
+const ProtectedRoute: React.FC<Properties> = ({
 	component,
 	redirectTo = AppRoute.SIGN_IN,
 }) => {
--- a/apps/frontend/src/libs/components/protected-route/protected-route.tsx
+++ b/apps/frontend/src/libs/components/protected-route/protected-route.tsx
@@ -1,5 +1,7 @@
 import type { FC, ReactNode } from "react";
+
 import { Navigate } from "react-router-dom";
+
 import { AppRoute } from "~/libs/enums/app-route.enum.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 import { type ValueOf } from "~/libs/types/types.js";
@@ -18,7 +20,7 @@ const ProtectedRoute: FC<Properties> = ({
 	const hasUser = Boolean(user);
 
 	if (!hasUser) {
-		return <Navigate to={redirectTo} replace />;
+		return <Navigate replace to={redirectTo} />;
 	}
 
 	return <>{component}</>;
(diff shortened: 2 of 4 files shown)


apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
+import type { FC, ReactNode } from "react";
```
fshabanov, line 1: Use React.FC and React.ReactNode

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { Navigate } from "react-router-dom";
```
fshabanov, line 3: Export this from libs/components/components

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
fshabanov, line 5:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
...
+type Properties = {
+	component: ReactNode;
+	redirectTo?: ValueOf<typeof AppRoute>;
+};
+
+const ProtectedRoute: FC<Properties> = ({
+	component,
+	redirectTo = AppRoute.SIGN_IN,
+}) => {
+	const { user } = useAppSelector((state) => state.auth);
```
fshabanov, line 18: Check rule B27

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+	const hasUser = Boolean(user);
+
+	if (!hasUser) {
+		return <Navigate replace to={redirectTo} />;
```
fshabanov, line 23: Why do we need `replace` here?

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -7,10 +7,12 @@ import { signUp } from "./actions.js";
 
 type State = {
 	dataStatus: ValueOf<typeof DataStatus>;
+	user: null;
```
fshabanov, line 10: User can be other than null

fshabanov (changes requested):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
tarmiel: The exceptions are single-line arrow functions, which can be used without the required blocks of code

apps/frontend/src/libs/components/protected-route/protected-route.tsx
tarmiel: By using replace, you ensure that the history stack doesn't accumulate unnecessary entries, leading to a cleaner browsing experience and it allows to seamlessly navigate users through the flow without cluttering the history with intermediate steps.

apps/frontend/src/modules/auth/slices/auth.slice.ts
tarmiel: As I mentioned in the comment to this issue I mocked user as null, because we agreed with my colleague who's working on the related functionallity that he's responsible for adding proper type here

apps/frontend/src/libs/components/protected-route/protected-route.tsx
tarmiel: If you mean ProtectedRoute component, it's already been done that way

fshabanov: I mean Navigate component

apps/frontend/src/modules/auth/slices/auth.slice.ts
fshabanov: You can add the type for User on your side as well and use it here. Other people, who will be working on User type will be able to modify it. You can discuss with others a proper naming, file location and other points so that you have same files for User type

apps/frontend/src/libs/components/protected-route/protected-route.tsx
tarmiel: Are you sure? Firstly this is a library component and I don`t understand why it is important to do that, and secondly, I think I will get circular dependency if I place Navigate in libs/components/components and try to import it to a protected route component that also exported from there, no?

fshabanov: The reason we put library components in libs/components/components is that we don't import library components directly in a file, unless it won't be used anywhere else. In this case Navigate can be used in other places as well. If we import directly from the library in multiple files and in future we decide to change the library, we would need to change it in all files. When we export from the library in one file only it is much easier.
Regarding circular dependencies, it might cause an issue only when one of the imports is not yet initialized and is being used. In our case we won't have the issue here.

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -2,15 +2,18 @@ import { createSlice } from "@reduxjs/toolkit";
 
 import { DataStatus } from "~/libs/enums/enums.js";
 import { type ValueOf } from "~/libs/types/types.js";
+import { type UserGetAllItemResponseDto as User } from "~/modules/users/users.js";
```
what1s1ove, line 5:
```suggestion
import { type UserGetAllItemResponseDto } from "~/modules/users/users.js";
```

apps/frontend/src/index.tsx
```diff
@@ -19,7 +20,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 					{
 						children: [
 							{
-								element: "Root",
+								element: <ProtectedRoute component={"Root"} />,
```
what1s1ove, line 23:
```suggestion
								element: <ProtectedRoute component="Root" />,
```

what1s1ove (changes requested):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Properties = {
+	component: React.ReactNode;
+	redirectTo?: ValueOf<typeof AppRoute>;
+};
+
+const ProtectedRoute: React.FC<Properties> = ({
+	component,
+	redirectTo = AppRoute.SIGN_IN,
+}: Properties) => {
```
fshabanov, line 14:
```suggestion
}) => {
```
`React.FC<Properties>` already handles this

fshabanov (changes requested):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
tarmiel: No, it does not. There are lint errors with missing react/props and I looked through other components to find this solution with duplicating Properties

fshabanov: You're right, sorry for misunderstanding

what1s1ove: Right, this is a part of `eslint-react` config

what1s1ove (approved):

fshabanov (approved):

PR #49, opened 2024-02-12
feat: get authenticated user tm-16

ShadowOfTheSpace:
### Overview

- Added `UserInfoResponse` type
- Made `contentType` optional in `getHeaders` method and `HTTPApiOptions` type
- Added `getAuthenticatedUser` method to `AuthApi` class
- Added `getAuthenticatedUser` redux `action` and `reducer`
- Added fetching `user by token` in `localStorage` after reloading page

### DIFF
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -7,23 +7,31 @@ import {
 	useEffect,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
+import { StorageKey, storage } from "~/libs/modules/storage/storage.js";
 import { actions as userActions } from "~/modules/users/users.js";
+import { actions as authActions } from "~/modules/auth/auth.js";
 
 const App: React.FC = () => {
 	const { pathname } = useLocation();
 	const dispatch = useAppDispatch();
-	const { dataStatus, users } = useAppSelector(({ users }) => ({
+	const { dataStatus, users } = useAppSelector(({ users, auth }) => ({
 		dataStatus: users.dataStatus,
 		users: users.users,
+		user: auth.user,
 	}));
 
 	const isRoot = pathname === AppRoute.ROOT;
+
+	const hasToken = Boolean(storage.get(StorageKey.TOKEN));
 
 	useEffect(() => {
 		if (isRoot) {
 			void dispatch(userActions.loadAll());
 		}
-	}, [isRoot, dispatch]);
+		if (hasToken) {
+			void dispatch(authActions.getAuthenticatedUser());
+		}
+	}, [isRoot, hasToken, dispatch]);
 
 	return (
 		<>
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -8,16 +8,16 @@ import {
 	useLocation,
 } from "~/libs/hooks/hooks.js";
 import { StorageKey, storage } from "~/libs/modules/storage/storage.js";
-import { actions as userActions } from "~/modules/users/users.js";
 import { actions as authActions } from "~/modules/auth/auth.js";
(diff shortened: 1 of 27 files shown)


apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -7,23 +7,31 @@ import {
...
-	const { dataStatus, users } = useAppSelector(({ users }) => ({
+	const { dataStatus, users } = useAppSelector(({ auth, users }) => ({
 		dataStatus: users.dataStatus,
+		user: auth.user,
 		users: users.users,
 	}));
 
 	const isRoot = pathname === AppRoute.ROOT;
 
+	const hasToken = Boolean(storage.get(StorageKey.TOKEN));
```
fshabanov, line 25: Move this to the action

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -18,15 +18,27 @@ const initialState: State = {
...
-		});
-		builder.addCase(signUp.fulfilled, (state) => {
-			state.dataStatus = DataStatus.FULFILLED;
-		});
-		builder.addCase(signUp.rejected, (state) => {
-			state.dataStatus = DataStatus.REJECTED;
+		builder.addCase(getAuthenticatedUser.fulfilled, (state, action) => {
+			state.user = action.payload;
 		});
+		builder.addMatcher(
```
fshabanov, line 24: Use `addCase` for each case separately

packages/shared/src/modules/users/libs/types/user-info-response.type.ts
```diff
@@ -0,0 +1,8 @@
+type UserInfoResponse = {
```
fshabanov, line 1:
```suggestion
type UserGetAuthenticatedResponseDto = {
```

fshabanov (changes requested):

packages/shared/src/modules/users/libs/types/user-info-response.type.ts
what1s1ove: I am not sure that we need an additional type here. I think you can use `UserSignInResponseDto | UserSignUpResponseDto`

what1s1ove (changes requested):

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -7,13 +7,15 @@ import {
...
+import { actions as authActions } from "~/modules/auth/auth.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
 const App: React.FC = () => {
 	const { pathname } = useLocation();
 	const dispatch = useAppDispatch();
-	const { dataStatus, users } = useAppSelector(({ users }) => ({
+	const { dataStatus, users } = useAppSelector(({ auth, users }) => ({
 		dataStatus: users.dataStatus,
+		user: auth.user,
```
fshabanov, line 18: But you are not using it

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -18,4 +20,16 @@ const signUp = createAsyncThunk<
...
+const getAuthenticatedUser = createAsyncThunk<
+	UserAuthResponseDto | null,
+	undefined,
+	AsyncThunkConfig
+>(`${sliceName}/get-authenticated-user`, (_payload, { extra }) => {
+	const { authApi, storage } = extra;
+
+	const hasToken = Boolean(storage.get(StorageKey.TOKEN));
+
+	return hasToken ? authApi.getAuthenticatedUser() : null;
```
fshabanov, line 32:
```suggestion
  if (!hasToken) {
    return null;
  }
  
	return authApi.getAuthenticatedUser();
```

fshabanov (changes requested):

fshabanov (approved):

PR #67, opened 2024-02-13
feat: authorization token tm-13

perekotypole:


apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -76,6 +76,16 @@ class BaseConfig implements Config {
 					format: Number,
 				},
 			},
+			JWT: {
+				ALGORITHM: "HS256",
+				EXPIRES_IN: "24h",
```
fshabanov, line 81: Move algorithm and expires in values to env

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -1,9 +1,29 @@
...
-		// TODO implement. This is just stub
-		return `token from ${JSON.stringify(data)}`;
+	private algorithm: string;
+	private expiresIn: string;
+	private secret: Uint8Array;
+
+	public constructor(config: Config) {
+		this.algorithm = config.ENV.JWT.ALGORITHM;
+		this.expiresIn = config.ENV.JWT.EXPIRES_IN;
+		this.secret = new TextEncoder().encode(config.ENV.JWT.SECRET);
```
fshabanov, line 15:
```suggestion
		this.secret = config.ENV.JWT.SECRET;
```

fshabanov (changes requested):

apps/backend/src/libs/modules/token/base-token.module.ts
tarmiel: in jose SignJWT.sign(key: Uint8Array | KeyLike) string won`t work

fshabanov: Then try `Buffer.from(config.ENV.SECRET, "utf8");`

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -14,10 +14,14 @@ const signUp = createAsyncThunk<
...
-	const { authApi } = extra;
+>(`${sliceName}/sign-up`, async (registerPayload, { extra }) => {
+	const { authApi, storage } = extra;
+
+	const { token, user } = await authApi.signUp(registerPayload);
+
+	await storage.set(StorageKey.TOKEN, token);
 
-	return authApi.signUp(registerPayload);
+	return { token, user };
```
what1s1ove, line 24:
```suggestion
	return user;
```
not sure that we need to pass a token somewhere

apps/backend/src/libs/modules/token/libs/types/token-payload.type.ts
```diff
@@ -0,0 +1,5 @@
+type TokenPayload = {
```
what1s1ove, line 1: i think we need move it to shared folder, since then we will need it on the frontend too

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -1,9 +1,29 @@
-import { Token } from "./libs/types/types.js";
+import { JWTVerifyResult, SignJWT, jwtVerify } from "jose";
+
+import { type Config } from "~/libs/modules/config/config.js";
+
+import type { Token, TokenPayload } from "./libs/types/types.js";
```
what1s1ove, line 5:
```suggestion
import {  type Token, type TokenPayload } from "./libs/types/types.js";
```

what1s1ove (changes requested):

apps/backend/src/libs/modules/token/base-token.module.ts
what1s1ove: merge main into your branch

what1s1ove (approved):

PR #80, opened 2024-02-14
feat: add udemy module tm-65

KrasKa2:


apps/backend/.env.example
```diff
@@ -14,6 +14,12 @@ DB_POOL_MIN=2
 DB_POOL_MAX=10
 
 #
+# UDEMY
+#
+UDEMY_CLIENT_ID=moCny32Y4SCu61N5xLMIBoMZhgtLORa7YQw3kKUW
+UDEMY_CLIENT_SECRET=xl6v73HlCkd67p79jpka5DwQvHgqbgmvSfggoOYmzFFz8rGsiGAVTYMftluCrBthDeicjT5oYXxBadCltSGmMr0d2JJvK1lUaOPKvMHfRwCAOT5AD0xCoO3SO5Izvpfg
```
fshabanov, line 20: Remove these values from the repo, it is not secure to put them here

apps/backend/src/libs/modules/http/base-http.module.ts
```diff
@@ -0,0 +1,84 @@
...
+	ContentType,
+	HTTPCode,
+	HTTPError,
+	HTTPHeader,
+	type HTTPOptions,
+} from "./libs/enums/enums.js";
+import { queryString } from "./libs/helpers/helpers.js";
+import { Http } from "./libs/types/types.js";
+
+// TODO there is already exists HTTPMethod type in shared dir
```
fshabanov, line 13: Use the type from the shared dir

apps/backend/src/libs/modules/http/base-http.module.ts
```diff
@@ -0,0 +1,84 @@
...
+	private parseJSON = <T>(response: Response): Promise<T> => {
+		return response.json() as Promise<T>;
+	};
+
+	private throwError = (error: Error): never => {
+		throw error;
+	};
+
+	private getHeaders(headers: Headers): Headers {
+		headers.append(HTTPHeader.CONTENT_TYPE, ContentType.JSON.toString());
```
fshabanov, line 58:
```suggestion
		headers.append(HTTPHeader.CONTENT_TYPE, ContentType.JSON);
```

apps/backend/src/libs/modules/http/libs/helpers/helpers.ts
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable unicorn/prefer-export-from */
+import queryString from "query-string";
```
fshabanov, line 2: Use `URLSearchParams`

apps/backend/src/libs/modules/udemy/udemy.ts
```diff
@@ -0,0 +1,29 @@
+import { config } from "~/libs/modules/config/config.js";
+
+import { http } from "../http/http.js";
+import { BasseUdemy } from "./base-udemy.module.js";
+import { CourseCurriculumFields } from "./libs/enums/course-curriculum-fields.enum.js";
```
fshabanov, line 5:
```suggestion
import { CourseCurriculumFields } from "./libs/enums/enums.js";
```

fshabanov (changes requested):

apps/backend/src/libs/modules/http/libs/types/http.type.ts
```diff
@@ -0,0 +1,7 @@
+import { HTTPOptions } from "../enums/enums.js";
+
+type Http = {
```
what1s1ove, line 3: you have this type on frontend at least. so move all shared things to shared module

apps/backend/src/libs/modules/udemy/base-udemy.module.ts
```diff
@@ -0,0 +1,105 @@
...
+	clientSecret: string;
+	fields: {
+		course: Record<string, string>;
+		courseCurriculum: Record<string, string>;
+		courseDetails: Record<string, string>;
+	};
+	http: Http;
+};
+
+class BasseUdemy<C, CD, CC> implements Udemy<C, CD, CC> {
```
what1s1ove, line 19:
```suggestion
class BaseUdemy<C, CD, CC> implements Udemy<C, CD, CC> {
```

what1s1ove, line 19:
```suggestion
class Udemy {
```
you do not need a type for implementation

just describe all methods that you need and use this class a type to pass to other services

what1s1ove (changes requested):

apps/backend/src/libs/modules/udemy/base-udemy.module.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	private getOptions(query: Record<string, unknown>): Partial<HTTPOptions> {
+		const headers = new Headers();
+		headers.append(HTTPHeader.AUTHORIZATION, this.authHeader);
+
+		return { headers, query };
+	}
+
+	public async getCourseCurriculums(id: number): Promise<CC | null> {
```
what1s1ove, line 54: move now you need only one method - getCourseBySlug or something like this

what1s1ove (changes requested):

apps/backend/.env.example
KrasKa2: Ups! I am sorry!

apps/backend/src/libs/modules/http/base-http.module.ts
KrasKa2: Точно)))

apps/backend/src/libs/modules/http/libs/types/http.type.ts
KrasKa2: @fshabanov , can I replace your variant of `HTTP` type with [mine](https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/blob/65-feat-udemy-module/apps/backend/src/libs/modules/http/libs/types/http.type.ts) (only in upper case) ? Because I need parameter <T> for load method, and Partial for options.

KrasKa2: @what1s1ove but I need another type (with parameter<T>). Maybe, I can rename it to HTTPApi?

fshabanov: This will result in unnecessary changes for frontend, I think the current type should be enough

KrasKa2: @fshabanov , I mean, I want to leave Http type on backend (and rename it to HTTPApi)

KrasKa2: I think, I understand...

PR #86, opened 2024-02-14
fix: layout for sign in page and validation messages tm-10

Ann-Maksay:
Auto-fill styles:

image

Validation error:
image


apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -29,6 +28,15 @@
...
 	background-color: var(--color-input);
+	border: 1px solid var(--color-input);
+}
+
+.dark:-webkit-autofill,
+.dark:-webkit-autofill:hover,
+.dark:-webkit-autofill:focus,
+.dark:-webkit-autofill:active {
+	border: 1px solid var(--color-background);
+	transition: background-color 1s 5000s;
```
Ann-Maksay, line 39: Is there no better way to deal with some browsers' styles that include `!important` and don't allow to change the background when auto-filling?

Vfivew (approved):

GEOFARL (approved):

ShadowOfTheSpace (approved):

MappariChan (approved):

apps/frontend/src/libs/components/input/styles.module.css
what1s1ove: it's okay. if QA will return this ticket, please, say about it on the daily. I will answer their questions

what1s1ove (approved):

PR #93, opened 2024-02-15
fix: sign up client tm-12

Sytnikovakatya:
1. Update styles to fix form layout when errors appear
2. Update validation schema, rules, and messages
3. Fix logo 

Screenshot_4


apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -55,15 +56,15 @@ const Auth: React.FC = () => {
 					styles[pathname.replace("/", "")],
 				)}
 			>
-				<h1 className={styles["logo-wrapper"]}>
+				<div className={styles["logo-wrapper"]}>
 					<img alt="TrackMates logo" className={styles["logo"]} src={logo} />
```
what1s1ove, line 60: we need to use `picture` element instead

package-lock.json
what1s1ove, line 1: this file should not be changed since you did not install/any packages

what1s1ove (changes requested):

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -1,27 +1,21 @@
 import { UserValidationRule } from "./user-validation-rule.enum.js";
 
 const UserValidationMessage = {
-	EMAIL_INVALID_FORMAT:
-		"Invalid email format. Please enter a valid email address.",
-	EMAIL_REQUIRE: "Email is required",
+	EMAIL_INVALID_FORMAT: "The email is invalid",
+	EMAIL_REQUIRE: "The field is required",
```
fshabanov, line 5: EMAIL_REQUIRE and FIELD_REQUIRE are now the same

fshabanov (changes requested):

package-lock.json
Sytnikovakatya: I don't now why it has changed. I've just merged main branch. I didn't install anything
What shall I do with it now?

Sytnikovakatya: Coulld I just delete it?

YourTeaChan (approved):

package-lock.json
ShadowOfTheSpace: I think, if you did not install or delete any packages, you should not commit `package-lock.json` while merging. So `package-lock.json` will not change on remote repo.

ShadowOfTheSpace (approved):

GEOFARL (approved):

package-lock.json
what1s1ove, line 1: this file should not be changed since you did not change anything in package.jsons

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -55,15 +57,15 @@ const Auth: React.FC = () => {
 					styles[pathname.replace("/", "")],
 				)}
 			>
-				<h1 className={styles["logo-wrapper"]}>
-					<img alt="TrackMates logo" className={styles["logo"]} src={logo} />
-					<img
+				<div className={styles["logo-wrapper"]}>
+					<Image alt="TrackMates logo" className={styles["logo"]} src={logo} />
```
what1s1ove, line 61: we have 3 images here. it's not that we want. please, use `picture` tag here instead

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture

what1s1ove (changes requested):

package-lock.json
Sytnikovakatya: Finally, I fixed it. Thanks a lot @ShadowOfTheSpace

Ann-Maksay (approved):

MappariChan (approved):

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -55,15 +57,18 @@ const Auth: React.FC = () => {
...
+					<picture>
+						<source media="(max-width: 480px)" srcSet={mobileLogo} />
+						<source media="(min-width: 481px)" srcSet={logo} />
+						<Image
+							alt="TrackMates logo"
+							className={`${styles["logo"]} ${styles["mobile-logo"]}`}
+							src={logo}
+						/>
+					</picture>
+					<Image alt="TrackMates" src={logoName} />
```
what1s1ove, line 70: why we need two images here? can we use only one here?

Sytnikovakatya: The second image we use for logo text ("TrackMates"). It was pointed by QA, because they wanted  pixel perfect logo like in Figma

fshabanov: I don't think that we should download this as a logo, because it's just a text
@what1s1ove what do you think?

Vfivew (approved):

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -55,15 +56,17 @@ const Auth: React.FC = () => {
...
-					/>
-					TrackMates
-				</h1>
+				<div className={styles["logo-wrapper"]}>
+					<picture>
+						<source media="(max-width: 480px)" srcSet={mobileLogo} />
+						<source media="(min-width: 481px)" srcSet={logo} />
+						<Image
+							alt="TrackMates logo"
+							className={`${styles["logo"]} ${styles["mobile-logo"]}`}
```
Oleksandra-Masiuk, line 65: there is already implemented clsx helper for this

Oleksandra-Masiuk (changes requested):

Oleksandra-Masiuk (approved):

what1s1ove (approved):

PR #97, opened 2024-02-16
build: continuous delivery tm-3

fshabanov:


apps/backend/package.json
```diff
@@ -2,6 +2,10 @@
 	"name": "backend",
 	"type": "module",
 	"version": "1.0.0",
+	"engines": {
+		"node": "18.x.x",
+		"npm": ">=9.x.x"
```
fshabanov, line 7: AWS used npm v9 with Node v18

package.json
```diff
@@ -8,15 +8,16 @@
...
+		"npm": ">=9.x.x"
 	},
 	"scripts": {
 		"git:hook:precommit": "lint-staged --cwd .",
 		"git:hook:commit": "npx @commitlint/cli --edit",
 		"build:shared": "npm run build -w shared",
 		"build:frontend": "npm run build -w frontend",
 		"build:backend": "npm run build -w backend",
-		"build": "npm run build --workspaces",
+		"build": "npm run build:shared && npm run build:backend && npm run build:frontend && npm run prepare-build",
```
fshabanov, line 19: `npm run build --workspaces` will not work because we have to build shared first. Backend and Frontend require this, otherwise they won't be built successfully

.github/workflows/cd.yml
```diff
@@ -0,0 +1,65 @@
...
+    steps:
+      - name: Code Checkout
+        uses: actions/checkout@v4
+
+      - name: Release
+        id: release
+        uses: google-github-actions/release-please-action@v3
+
+  deploy:
+    needs: release
```
what1s1ove, line 39: let's add deployment url here - https://github.com/what1s1ove/whatislove.dev/blob/main/.github/workflows/cd.yml#L111
it will be hardcoded for now since we have only one ENV

what1s1ove (approved):

what1s1ove (approved):

PR #100, opened 2024-02-16
fix: "User not found" error when an authorized user is trying to navigate to the root page tm-95

ShadowOfTheSpace:


what1s1ove (approved):

apps/frontend/src/libs/helpers/check-if-show-loader/check-if-show-loader.helper.ts
```diff
@@ -0,0 +1,8 @@
+import { DataStatus } from "~/libs/enums/enums.js";
+import { ValueOf } from "~/libs/types/value-of.type.js";
```
GEOFARL, line 2:
```suggestion
import { ValueOf } from "~/libs/types/types.js";
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -26,6 +30,10 @@ const App: React.FC = () => {
 		void dispatch(authActions.getAuthenticatedUser());
 	}, [isRoot, dispatch]);
 
+	if (checkIfShowLoader(authDataStatus)) {
```
fshabanov, line 33: We don't need a helper for that, it is only used in one file

fshabanov (changes requested):

GEOFARL (approved):

YourTeaChan (approved):

Vfivew (approved):

Oleksandra-Masiuk (approved):

PR #118, opened 2024-02-18
fix: sign page scroll tm-115

Ann-Maksay:
image
image


Vfivew (approved):

.prettierignore
```diff
@@ -0,0 +1 @@
+*.md
```
what1s1ove, line 1: please pull main into your branch

what1s1ove (changes requested):

apps/frontend/src/pages/auth/styles.module.css
```diff
@@ -30,7 +30,7 @@
 	align-items: center;
 	justify-content: center;
 	width: 454px;
-	min-height: 685px;
+	height: 100%;
```
MappariChan, line 33:
```suggestion
	height: 100vh;
```
I think if you write min-height in 100vh you should also use this measurements for height attribute

MappariChan (approved):

fshabanov (approved):

PR #125, opened 2024-02-20
fix(frontend): The Overview page doesn't occupy 100% of the screen height tm-121

GEOFARL:
image


MappariChan (approved):

what1s1ove (approved):

apps/frontend/src/libs/components/auth-wrapper/styles.module.css
```diff
@@ -1,6 +1,6 @@
 .page-layout {
 	display: flex;
-	height: 100%;
+	min-height: 100vh;
```
what1s1ove, line 3: 100vh should be only on the body
for the rest of element use flex-grow

what1s1ove (changes requested):

Sytnikovakatya (approved):

what1s1ove (approved):

PR #134, opened 2024-02-20
feat(backend/shared): add AI integration tm-43

YourTeaChan:


YourTeaChan:
Please check only related packages to my PR (open-ai and recommended-courses)

apps/backend/src/modules/open-ai/libs/enums/open-ai-properties.enum.ts
```diff
@@ -0,0 +1,8 @@
+const OpenAiProperties = {
+	MODEL: "gpt-3.5-turbo-0125",
```
what1s1ove, line 2:
```suggestion
```
move it to env as well

apps/backend/src/modules/open-ai/libs/enums/open-ai-properties.enum.ts
```diff
@@ -0,0 +1,8 @@
...
+	RESPONSE_FORMAT: { type: "text" },
```
what1s1ove, line 3: all these properties should be default. so just add them to AI call

apps/backend/src/modules/open-ai/libs/enums/prompt.enum.ts
```diff
@@ -0,0 +1,9 @@
+import { PromptInputValue } from "./prompt-input-value.enum.js";
+import { PromptOutputExample } from "./prompt-output-example.enum.js";
+import { PromptOutputValue } from "./prompt-output-value.js";
+
+const Prompt = {
```
what1s1ove, line 5: i thing you do not need all this prompt is separate file. just add it to the method that will call AI service

apps/backend/src/modules/open-ai/open-ai.controller.ts
```diff
@@ -0,0 +1,113 @@
+import { APIPath, HTTPCode } from "~/libs/enums/enums.js";
```
what1s1ove, line 1: open ai will not have a separate controller and etc. AI should be just a service with one method for now `call` or similar. inject this service in source module, all courses functionality you need to describe in course module

what1s1ove (changes requested):

apps/backend/src/modules/open-ai/libs/enums/prompt.enum.ts
YourTeaChan: Do you mean I don't need all prompt enums and I just should add a large string into the AI call method (`getResponseFromOpenAi`)? Will it look readable?
Or do you mean there must be only a `Prompt` enum, which will contain only a string?

apps/backend/src/modules/open-ai/open-ai.controller.ts
YourTeaChan: Do you mean I need to keep `OpenAiService` with only one method named `call`, which will call `openAi.chat.completions.create` method? And all other logic I should move to `CourseService` (mapping to another type, etc)?
We discussed in the one of previous meetings about a separate block with recommended courses and why do we need it. But now we will sort all courses that user will search and will display it sorted (and we will not have a recommended courses block now)?

Sytnikovakatya (approved):

apps/backend/src/modules/open-ai/open-ai.service.ts
```diff
@@ -0,0 +1,41 @@
...
+type Constructor = {
+	model: string;
+	openAi: OpenAI;
+};
+
+class OpenAiService {
+	private model: string;
+	private openAi: OpenAI;
+
+	constructor({ model, openAi }: Constructor) {
```
fshabanov, line 15:
```suggestion
	constructor({ model, openAI }: Constructor) {
```
Change everywhere

fshabanov (changes requested):

GEOFARL (approved):

ShadowOfTheSpace (approved):

apps/backend/src/modules/open-ai/open-ai.service.ts
```diff
@@ -0,0 +1,41 @@
+import { type OpenAI } from "~/libs/modules/open-ai/open-ai.js";
```
what1s1ove, line 1: we do not want to have a open-ai as an app module (users, courses, etc). it should be placed in `apps/backend/src/libs/modules/open-ai`

apps/backend/src/modules/open-ai/libs/enums/prompt.enum.ts
what1s1ove: > Do you mean I don't need all prompt enums and I just should add a large string into the AI call method (getResponseFromOpenAi)? Will it look readable?

Yes, let's do this way

but you need to inject `open-ai` service inside `courses` service and add this string while calling `open-ai` service from there

what1s1ove (changes requested):

MappariChan (approved):

apps/backend/src/libs/modules/open-ai/libs/enums/open-ai-error-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const OpenAIErrorMessage = {
+	WRONG_RESPONSE: "Wrong response from OpenAI",
```
fshabanov, line 2:
```suggestion
	WRONG_RESPONSE: "Wrong response from OpenAI.",
```
Add dot to have the same style everywhere

apps/backend/src/libs/modules/open-ai/libs/exceptions/open-ai-error.exception.ts
```diff
@@ -0,0 +1,14 @@
+import { ApplicationError } from "~/libs/exceptions/exceptions.js";
+
+type Constructor = {
+	cause?: unknown;
+	message: string;
```
fshabanov, line 5: Should we add status as well?

fshabanov (changes requested):

apps/backend/src/libs/modules/open-ai/libs/exceptions/open-ai-error.exception.ts
YourTeaChan: `OpenAIError` extends `ApplicationError`, not `HTTPError` (we have status only in `HTTPError` and its children). Or I need to extend `HTTPError` instead of `ApplicationError`?

Vfivew (approved):

.github/workflows/ci-pr.yml
```diff
@@ -88,7 +88,7 @@ jobs:
           path: |
```
what1s1ove, line 88: please, merge main branch inside your. we should see only your changes

apps/backend/src/modules/courses/course.service.ts
```diff
@@ -181,8 +202,9 @@ class CourseService {
 		let courses = vendorsCourses.flat();
 
 		courses = await this.filterCourses(courses, userId);
+		const sortedCourses = await this.sortCoursesByOpenAI(courses);
 
-		return { courses };
+		return { courses: sortedCourses.filter(Boolean) };
```
what1s1ove, line 207: it should be as a separate call on frontend to get AI courses list. you do not need to change this method

what1s1ove (changes requested):

apps/backend/src/modules/courses/course.service.ts
YourTeaChan: At one of previous daily calls you said that I need to remove controller and leave only a openAI module with `call` method. So, how can we make it without openAI controller?
> a separate call on frontend to get AI courses list

Or I need to create separate endpoint at `Course controller`? But if I will do that, I will duplicate the `findAllByVendor` method.

KrasKa2 (changes requested):

ShadowOfTheSpace (approved):

GEOFARL (approved):

KrasKa2 (approved):

apps/backend/src/modules/courses/course.service.ts
```diff
@@ -186,6 +191,28 @@ class CourseService {
...
 
+	public async getRecommendedCoursesByAI(parameters: {
+		search: string;
+		userId: number;
+		vendorsKey: string | undefined;
+	}): Promise<CoursesResponseDto> {
+		const { courses } = await this.findAllByVendors(parameters);
+
+		const prompt =
+			"Given a list of courses in JSON format. Analyze these courses by all fields. Return an array with the same length as the input array. Returned array must contains indexes of sorted courses based on their overall quality and interest level, from the most recommended courses to the least. The answer should be without any explanation in the format: [index]. The index starts with 0. For example: the input data to analyze is [{0}, {1}, {2}]. After analyzing, the output should be like: [1, 2, 0]. Please make sure that the returned array has the same length as the input array.";
```
fshabanov, line 202:
```suggestion
			"Given a list of courses in JSON format. Analyze these courses by all fields. Return an array with the same length as the input array. Returned array must contain indexes of sorted courses based on their overall quality and interest level, from the most recommended courses to the least. The answer should be without any explanation in the format: [index]. The index starts with 0. For example: the input data to analyze is [{0}, {1}, {2}]. After analyzing, the output should be like: [1, 2, 0]. Please make sure that the returned array has the same length as the input array.";
```

fshabanov (changes requested):

what1s1ove (approved):

PR #135, opened 2024-02-21
feat(backend/shared): chat functionality tm-111

ShadowOfTheSpace:


packages/shared/src/modules/chat/chat.ts
```diff
@@ -0,0 +1,12 @@
...
+	type ChatGetAllResponseDto,
+	type ChatItemResponseDto,
+	type MessageGetAllResponseDto,
+	type MessageResponseDto,
+	type MessageSendRequestDto,
+} from "./libs/types/types.js";
+export {
+	chatParameters as chatParametersValidationSchema,
+	messageSend as messageSendValidationSchema,
+} from "./libs/validation-schemas/validation-schemas.js";
```
Ann-Maksay, line 12: Wouldn't it be better to keep the same module names in all directories?
Either `chat` or `chat-message` for both shared and backend (and frontend in the future).

apps/backend/src/modules/chat-message/chat-message.controller.ts
```diff
@@ -0,0 +1,214 @@
...
+	 *                properties:
+	 *                  chats:
+	 *                    type: array
+	 *                    items:
+	 *                      type: object
+	 *                      $ref: "#/components/schemas/Chat"
+	 */
+	private async findAllChatsByUserId({
+		user: { id },
+	}: APIHandlerOptions<{
```
GEOFARL, line 125: I was asked to not use such destructuring and to access id on user obj, as user.id

apps/backend/src/modules/chat-message/chat-message.service.ts
```diff
@@ -0,0 +1,101 @@
...
+import { ChatMessageEntity } from "./chat-message.entity.js";
+import { type ChatMessageRepository } from "./chat-message.repository.js";
+import {
+	type ChatGetAllResponseDto,
+	type MessageGetAllResponseDto,
+	type MessageResponseDto,
+	type MessageSendRequestDto,
+} from "./libs/types/types.js";
+
+const ARRAY_EMPTY_LENGTH = 0;
```
GEOFARL, line 14: probably, to constants

Sytnikovakatya (approved):

packages/shared/src/modules/chat/libs/types/chat-item-response-dto.type.ts
```diff
@@ -0,0 +1,11 @@
+import { type UserAuthResponseDto } from "../../../users/users.js";
```
MappariChan, line 1:
```suggestion
import { type UserAuthResponseDto } from "~/modules/users/users.js";
```

GEOFARL: > 

If I'm not mistaken, we cannot do that in packages folder

ShadowOfTheSpace: > 

We can't do it in `package` folder. We have import aliases only for `apps` folder

apps/backend/src/modules/chat-message/chat-message.repository.ts
```diff
@@ -0,0 +1,235 @@
...
+				passwordSalt: chatMessage.senderUser.passwordSalt,
+				updatedAt: chatMessage.senderUser.updatedAt,
+			}),
+			status: chatMessage.status,
+			updatedAt: chatMessage.updatedAt,
+		});
+	}
+
+	public delete(): Promise<boolean> {
+		return Promise.resolve(true);
```
what1s1ove, line 83: implement all methods

apps/backend/src/db/migrations/20240218085611_add_chat_messages_table.ts
```diff
@@ -0,0 +1,52 @@
+import { type Knex } from "knex";
+
+import { MessageStatus } from "~/modules/chat-message/libs/enums/enums.js";
```
what1s1ove, line 3:
```suggestion
```
describe this enum in this file. ask on our daily why 🙂

packages/shared/src/modules/chat-message/libs/types/message-get-all-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type MessageResponseDto } from "./message-response-dto.type.js";
+
+type MessageGetAllResponseDto = {
+	messages: MessageResponseDto[];
```
what1s1ove, line 4:
```suggestion
	items: MessageResponseDto[];
```
soon we will unify all `GetAll` types

what1s1ove (changes requested):

apps/backend/src/modules/chat-message/chat-message.repository.ts
```diff
@@ -0,0 +1,235 @@
...
+	public find(): Promise<ChatMessageEntity | null> {
+		return Promise.resolve(null);
+	}
+
+	public findAll(): Promise<ChatMessageEntity[]> {
+		return Promise.resolve([]);
+	}
+
+	public async findAllChatsByUserId(userId: number) {
+		const chats = await this.chatMessageModel
```
fshabanov, line 95: Can we use `distinct` instead of all of this? Find chats by user id with distinct chatId and order them as you need

apps/backend/src/modules/chat-message/chat-message.service.ts
```diff
@@ -0,0 +1,100 @@
...
+		return {
+			chats: chats.map((chat) => chat.toObject()),
+		};
+	}
+
+	public async findAllMessagesByChatId(
+		chatId: string,
+		userId: number,
+	): Promise<MessageGetAllResponseDto> {
+		const userIsChatOwner =
```
fshabanov, line 40:
```suggestion
		const isUserChatOwner =
```

packages/shared/src/modules/chat-message/libs/enums/chat-message-api-path.enum.ts
```diff
@@ -0,0 +1,6 @@
+const ChatMessageApiPath = {
+	CHAT_ID: "/:chatId",
```
fshabanov, line 2:
```suggestion
	$CHAT_ID: "/:chatId",
```

packages/shared/src/modules/chat-message/libs/enums/chat-validation-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const ChatValidationMessage = {
+	CHAT_ID_INVALID_TYPE: "The chatId is incorrect uuid",
```
fshabanov, line 2:
```suggestion
	CHAT_ID_INVALID_TYPE: "The chatId is incorrect uuid.",
```

fshabanov (changes requested):

apps/backend/src/db/migrations/20240218085611_add_chat_messages_table.ts
ShadowOfTheSpace: @what1s1ove Probably I understand what you mean...
But how we could use it outside of migration file? We should do not use that enum at all and use type string instead? Or we need to have separate enums in migration file and in shared folder?

packages/shared/src/modules/chat/libs/types/chat-item-response-dto.type.ts
MappariChan: Oh, okay, my bad, sorry

apps/backend/src/modules/chat-message/chat-message.model.ts
```diff
@@ -0,0 +1,61 @@
...
+				relation: Model.BelongsToOneRelation,
+			},
+			selfChatMessages: {
+				join: {
+					from: `${DatabaseTableName.CHAT_MESSAGES}.chatId`,
+					to: `${DatabaseTableName.CHAT_MESSAGES}.chatId`,
+				},
+				modelClass: ChatMessageModel,
+				relation: Model.HasManyRelation,
+			},
```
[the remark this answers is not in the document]
ShadowOfTheSpace: I need it for creating relatedQuery to the same table.

I need to find a number of UNREAD messages for each user chat. So to make it in one db query I need that strange relation. It was the only way that I found to achieve a requested functionality in the same query while finding user's chats. 

If we could have separated "Chat" and "Message" tables, it would be much easier to achieve that (like it was in thread-js with post reactions counting for example), but right now chats and messages are contained in one table.
@fshabanov What do you think about that? Is another way exists?

apps/backend/src/modules/chat-message/chat-message.repository.ts
ShadowOfTheSpace: Short answer is no I think

Firstly if you have `distinct on (chat_id)`, you have to specify order by columns in the same order, as in distinct statement

So I cannot simply do something like that
`select distinct (chat_id) from chat_messages order by created_at`
because it throws error that I described before

If I will do that like postgres wants 
`select distinct on (chat_id) * from chat_messages order by chat_id, created_at desc`
I will get all chats, but they will be `sorted by chat_id` and only after that `by created_at`. So we cannot get chats sorted by last message in it with that approach
So for that reason I created `CTE`, which I use to sort user's chats properly

Additionally as I described in your previous comment I need to get not only user's chats, but I need to count unread messages in every chat, so I need subquery for that (that one with strange relation)

I think I cannot find better approach to achieve requested behaviour (for now)

GEOFARL (approved):

MappariChan (approved):

apps/backend/src/modules/chat-message/chat-message.model.ts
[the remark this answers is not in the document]
fshabanov: @ShadowOfTheSpace try to do something like this. You will not need `selfChatMessages` anymore
image

GEOFARL (approved):

apps/backend/src/modules/chats/chat.model.ts
```diff
@@ -0,0 +1,66 @@
+import { Model, type QueryBuilder, type RelationMappings } from "objection";
+
+import {
+	AbstractModel,
+	DatabaseTableName,
+} from "~/libs/modules/database/database.js";
+import { ChatMessageModel } from "~/modules/chat-messages/chat-message.model.js";
```
GEOFARL, line 7:
```suggestion
import { ChatMessageModel } from "~/modules/chat-messages/chat-messages.js";
```

apps/backend/src/modules/chats/chat.repository.ts
```diff
@@ -0,0 +1,307 @@
+import { type Repository } from "~/libs/types/types.js";
+import { ChatMessageEntity } from "~/modules/chat-messages/chat-message.entity.js";
+import { type ChatMessageModel } from "~/modules/chat-messages/chat-message.model.js";
```
GEOFARL, line 3:
```suggestion
import { ChatMessageEntity } from "~/modules/chat-messages/chat-messages.js";
import { type ChatMessageModel } from "~/modules/chat-messages/chat-messages.js";
```

apps/backend/src/modules/chats/chat.repository.ts
```diff
@@ -0,0 +1,307 @@
...
+import { UserEntity } from "~/modules/users/users.js";
+
+import { MessageStatus } from "../chat-messages/chat-message.js";
```
GEOFARL, line 6:
```suggestion
import { MessageStatus } from "~/modules/chat-messages/chat-messages.js";
```

apps/backend/src/modules/chats/chat.service.ts
```diff
@@ -0,0 +1,127 @@
+import { ExceptionMessage, HTTPCode } from "~/libs/enums/enums.js";
+import { HTTPError } from "~/libs/modules/http/http.js";
+import { type Service } from "~/libs/types/service.type.js";
```
GEOFARL, line 3:
```suggestion
import { type Service } from "~/libs/types/types.js";
```

GEOFARL: add it to other import types

packages/shared/src/modules/chat-messages/libs/types/chat-message-item-response-dto.type.ts
```diff
@@ -0,0 +1,15 @@
+import { type ValueOf } from "../../../../libs/types/value-of.type.js";
```
GEOFARL, line 1:
```suggestion
import { type ValueOf } from "../../../../libs/types/types.js";
```

packages/shared/src/modules/chat-messages/libs/types/chat-message-item-response-dto.type.ts
```diff
@@ -0,0 +1,15 @@
...
+import { type UserAuthResponseDto } from "../../../users/users.js";
+import { type MessageStatus } from "../enums/message-status.enum.js";
```
GEOFARL, line 3:
```suggestion
import { type MessageStatus } from "../enums/enums.js";
```

apps/backend/src/libs/types/service.type.ts
```diff
@@ -1,8 +1,8 @@
 type Service<T = unknown> = {
 	create(payload: unknown): Promise<T>;
-	delete(id: number): Promise<boolean>;
+	delete(id: unknown): Promise<boolean>;
```
what1s1ove, line 3: it should be number

apps/backend/src/modules/chats/chat.service.ts
```diff
@@ -0,0 +1,187 @@
...
+		if (userId !== firstUser.id && userId !== secondUser.id) {
+			throw new ChatError({
+				message: ExceptionMessage.CHAT_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		void (await this.chatRepository.delete(id));
+
+		return true;
```
what1s1ove, line 100:
```suggestion
	const deletedRowsCount = await this.chatRepository.delete(id);

	return Boolean(deletedRowsCount);
```

apps/backend/src/modules/chats/chat.service.ts
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	public async update(
+		id: number,
+		{
+			firstUserId,
+			secondUserId,
+		}: { firstUserId: number; secondUserId: number },
+	): Promise<ChatResponseDto> {
+		void (await this.find(id));
```
what1s1ove, line 163: add it to variable

apps/backend/src/modules/friends/friend.repository.ts
```diff
@@ -121,6 +121,23 @@ class FriendRepository implements Repository<UserEntity> {
 		});
 	}
 
+	public async getIsMutualFollowersByIds(
```
what1s1ove, line 124:
```suggestion
	public async checkIsMutualFollowersByIds(
```
methods that return boolean should start with `check<is | has>`

what1s1ove (changes requested):

apps/backend/src/modules/chat-messages/chat-message.controller.ts
```diff
@@ -0,0 +1,337 @@
...
+						body: ChatMessageUpdateRequestDto;
+						params: Record<"messageId", number>;
+						user: UserAuthResponseDto;
+					}>,
+				);
+			},
+			method: "PATCH",
+			path: ChatMessagesApiPath.$MESSAGE_ID,
+			validation: {
+				body: chatMessageUpdateValidationSchema,
```
fshabanov, line 97: Add `params` validation

apps/backend/src/modules/chat-messages/chat-message.repository.ts
```diff
@@ -0,0 +1,167 @@
...
+	public async findAll(userId: number): Promise<ChatMessageEntity[]> {
+		const messagesByUserId = await this.chatMessageModel
+			.query()
+			.where({ senderUserId: userId })
+			.withGraphFetched(
+				`${RelationName.SENDER_USER}.${RelationName.USER_DETAILS}`,
+			)
+			.execute();
+
+		return messagesByUserId.map((messageByUserId) =>
```
fshabanov, line 105: B27

apps/backend/src/modules/friends/friend.repository.ts
```diff
@@ -121,6 +121,23 @@ class FriendRepository implements Repository<UserEntity> {
...
+		firstUserId: number,
+		secondUserId: number,
+	): Promise<boolean> {
+		const [firstUserSubscription, secondUserSubscription] = await this.userModel
+			.query()
+			.from(DatabaseTableName.FRIENDS)
+			.where({ follower_id: firstUserId, following_id: secondUserId })
```
fshabanov, line 131:
```suggestion
			.where({ followerId: firstUserId, followingId: secondUserId })
```

fshabanov (changes requested):

apps/backend/src/modules/chats/chat.entity.ts
```diff
@@ -0,0 +1,208 @@
+import { type Entity } from "~/libs/types/types.js";
+import { type ChatMessageEntity } from "~/modules/chat-messages/chat-message.js";
```
fshabanov, line 2:
```suggestion
import { type ChatMessageEntity } from "~/modules/chat-messages/chat-messages.js";
```

fshabanov (changes requested):

what1s1ove (approved):

PR #140, opened 2024-02-21
fix(frontend): Create Friend component tm-82

GEOFARL:
No more square shadows
image
No more white square bottom left corner when hovering the component
image
The "Follow" button appearance should change according to the design (when hovering)
image
The first and last name font should be Roboto
image
The font of the "Follow" label should be Roboto
image
No white outline


YourTeaChan (approved):

apps/frontend/src/pages/friends/components/friend/styles.module.css
```diff
@@ -2,19 +2,21 @@
 	display: flex;
 	gap: 24px;
 	padding: 12px;
+	font-family: Roboto, sans-serif;
```
fshabanov, line 5: Maybe it should be a global font? So that we don't change this in every place. Clarify with QA

Sytnikovakatya (approved):

Ann-Maksay (approved):

apps/frontend/src/pages/friends/components/friend/styles.module.css
GEOFARL: QA team confirmed that it should be a global font

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -11,6 +11,7 @@ type Properties = {
 	color?: "primary" | "secondary";
 	hasVisuallyHiddenLabel?: boolean;
 	href?: ValueOf<typeof AppRoute>;
+	iconClassName?: string | undefined;
```
what1s1ove, line 14:
```suggestion
```
i not sure that you need it. fill of the icon should be as a `currentColor` - https://css-tricks.com/cascading-svg-fill-color/

text and icon should be always with the same color

what1s1ove (approved):

PR #154, opened 2024-02-23
fix(frontend): make roboto global font style tm-143

MappariChan:


apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -8,7 +8,7 @@ body {
 	min-width: 320px;
 	min-height: 100vh;
 	margin: 0;
-	font-family: Arial, sans-serif;
+	font-family: Roboto, sans-serif;
```
fshabanov, line 11: B14

fshabanov (changes requested):

apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -8,7 +8,7 @@ body {
 	min-width: 320px;
 	min-height: 100vh;
 	margin: 0;
-	font-family: Arial, sans-serif;
+	font-family: Roboto, Arial, sans-serif;
```
tarmiel, line 11: Roboto font isn't one of the default fonts in the browser. At this moment, we use Arial or sans-serif instead of Roboto. So, according to rule B13, you should install this font locally

KrasKa2 (approved):

Vfivew (approved):

ShadowOfTheSpace (approved):

YourTeaChan (approved):

apps/frontend/src/assets/css/scaffolding.css
MappariChan: If i correctly understand, i should create font localy using `@font-face`

apps/frontend/src/assets/css/fonts/roboto.css
tarmiel, line 1: At this moment our design has only Roboto regular(400), medium(500), semibold(600), bold(700). So, I'm not sure that we need other variants.

MappariChan: I think it will be better to leave it for now.

MappariChan: But maybe you are right, i will ask about it in the chat

fshabanov (approved):

fshabanov:
@MappariChan merge the main branch into yours

PR #167, opened 2024-02-25
fix(frontend): The pages don't occupy 100% of the screen height after changing the page width to less than 650px in the "Toggle device toolbar" in DevTools tm-136

GEOFARL:
image


GEOFARL:
I have also tried to fix an issue with new viewport units lvh, svh and dvh. However, that is not working

apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -8,6 +8,7 @@ body {
 	display: flex;
 	flex-direction: column;
 	min-width: 320px;
+	height: var(--doc-height);
```
what1s1ove, line 11:
```suggestion
	min-height: 100vh
```
and then to all wrappers you should add `flex-grow: 1`
you do not need to write any programatic functionality. css only

what1s1ove (changes requested):

apps/frontend/src/assets/css/scaffolding.css
GEOFARL: This has already been implemented, but the problem persists in the developer tools

GEOFARL: image

That is the body element with just min-height: 100vh

Sytnikovakatya (approved):

GEOFARL:
> Chrome Developer Tools are not part of the viewport object as far as CSS is concerned. You can use JS to size the page instead, but what you want cannot be accomplished by CSS alone. CSS vw & vh are calculated the same way as window.innerWidth and window.innerHeight.

[described here](https://stackoverflow.com/questions/57359440/how-to-enforce-100-vh-on-html-when-a-dev-tool-is-open-at-the-bottom)

I'm not sure whether that is actually an issue, becasue besides chrome dev tools everything works correctly
So probably this PR should be closed

what1s1ove:
Because of https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/167#issuecomment-1962861869

PR #169, opened 2024-02-25
fix(frontend): sidebar mobile pwa tm-139

Sytnikovakatya:
**Replace sidebar with bottom menu on PWA app**

Screenshot_10

### DIFF
--- a/apps/frontend/src/libs/components/sidebar/sidebar.tsx
+++ b/apps/frontend/src/libs/components/sidebar/sidebar.tsx
@@ -1,6 +1,6 @@
 import logo from "~/assets/img/website-logo.png";
 import { getValidClassNames } from "~/libs/helpers/helpers.js";
-import { useCallback, useState } from "~/libs/hooks/hooks.js";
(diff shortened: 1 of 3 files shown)


what1s1ove (approved):

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -43,13 +46,28 @@ const Sidebar: React.FC<Properties> = ({ menuItems }: Properties) => {
...
+									href === location.pathname ? styles["active"] : "",
+								)}
+								to={href}
+							>
+								{" "}
+								<span className={styles["menu-icon"]}>
+									<Icon name={icon} />
+								</span>
+								<span>{label}</span>
+							</Link>
```
MappariChan, line 69: Should we add both the button and link into sidebar component? If i correctly understand you wrote new styling for the link which is responsible for bottom tab, but i think you should rewrite existing styles for button.

Sytnikovakatya: Ok, I'll try it

PR #182, opened 2024-02-25
fix(frontend): sidebar only the button representing active page should be active tm-174

Vfivew:
I had to fix all the buttons.
Now there are three buttons, according to the layout, and a total of three styles.
btn


apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -60,10 +64,20 @@ const Sidebar: React.FC<Properties> = ({ menuItems }: Properties) => {
...
 							<Button
 								className={styles["menu-item"]}
+								color={
+									href === getFirstPathSegment(location.pathname)
+										? "primary"
+										: "secondary"
+								}
 								href={href}
 								iconName={icon}
 								key={label}
```
Ann-Maksay, line 74: This should have been spotted in another PR, but I guess it's worth a fix now. 
The key should be on the wrapper, not on the inner element.

image

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -31,13 +31,15 @@
 	color: var(--color-menu);
 	background-color: var(--main-white);
 	border-color: var(--main-white);
+	fill: var(--color-menu);
```
tarmiel, line 34: I think it would be better to set fill="currentColor" in svg to change such icons through color property in css

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -57,13 +61,22 @@ const Sidebar: React.FC<Properties> = ({ menuItems }: Properties) => {
 				</Link>
 				<nav className={styles["menu"]}>
 					{menuItems.map(({ href, icon, label }) => (
-						<>
+						<div key={label}>
 							<Button
 								className={styles["menu-item"]}
+								color={
```
what1s1ove, line 67: you need to use `activeClassName` prop for the react-router link component instead

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -31,13 +31,15 @@
...
 }
 
 .primary.outlined,
```
tarmiel, line 37: I'm not sure that we still need primary outlined variant

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -31,13 +31,15 @@
...
 .primary:hover {
 	color: var(--main-white);
 	background-color: var(--color-button);
 	border-color: var(--color-button);
+	fill: var(--main-white);
```
what1s1ove, line 42: the fill property should be added to the icon component and it should be `currentColor`, so by passing color to the button, the icon should inherit this color

https://css-tricks.com/cascading-svg-fill-color/

what1s1ove (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar.tsx
Vfivew: It seems to me it does not exist now ( i mean activeClassName ).
But I figured out how to do this more concisely.
https://github.com/tailwindlabs/tailwindcss/discussions/6729

what1s1ove: https://reactrouter.com/en/main/components/nav-link

GEOFARL (approved):

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -1,3 +1,4 @@
+import { useCallback } from "react";
```
fshabanov, line 1:
```suggestion
import { useCallback } from "~/libs/hooks/hooks.js";
```

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -57,35 +51,30 @@ const Sidebar: React.FC<Properties> = ({ menuItems }: Properties) => {
...
-							<Button
-								className={styles["menu-item"]}
-								href={href}
-								iconName={icon}
-								key={label}
-								label={label}
-							/>
+						<div key={label}>
 							<Link
+								activeClassName={styles["active"]}
```
MappariChan, line 56:
```suggestion
							<Link
							  key={label}
								activeClassName={styles["active"]}
```
I think you should omit this div and pass key directly into the Link component

fshabanov (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar.tsx
MappariChan, line 56:
```suggestion
							<Link
							  key={label}
								activeClassName={styles["active"]}
```
I think you should omit this div and pass key directly into the Link component

apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -62,16 +86,21 @@
...
+	.close .link-title {
+		display: none;
 	}
 
 	.burger-button {
 		z-index: var(--z-index-high);
+		color: var(--main-white);
+	}
+
+	.close .log-out-btn span {
```
GEOFARL, line 98: B10 Rule

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -57,35 +51,30 @@ const Sidebar: React.FC<Properties> = ({ menuItems }: Properties) => {
...
+								<Icon name={icon} />
+							</span>
+							<span className={styles["link-title"]}>{label}</span>
+						</Link>
 					))}
 				</nav>
 				<Button
-					className={styles["menu-item"]}
+					className={styles["log-out-btn"]}
+					hasVisuallyHiddenLabel={false}
```
tarmiel, line 73:
```suggestion

```

tarmiel (changes requested):

Sytnikovakatya (approved):

tarmiel (approved):

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -2,23 +2,37 @@ import { NavLink } from "react-router-dom";
...
 	to: ValueOf<typeof AppRoute>;
 };
 
 const Link: React.FC<Properties> = ({
+	activeClassName,
 	children,
 	className,
 	to,
 }: Properties) => {
+	const linkStyles = useCallback(
```
what1s1ove, line 23:
```suggestion
	const handleLinkStyles = useCallback(
```

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -8,34 +8,31 @@ import styles from "./styles.module.css";
...
 	className?: string | undefined;
-	color?: "primary" | "secondary";
 	hasVisuallyHiddenLabel?: boolean;
 	href?: ValueOf<typeof AppRoute>;
 	iconName?: IconName;
 	label: string;
 	onClick?: React.MouseEventHandler<HTMLButtonElement> | undefined;
 	size?: "regular" | "small";
-	style?: "filled" | "outlined";
+	style?: "plain" | "primary" | "secondary";
```
what1s1ove, line 17: not sure why you removed `color` prop
could you show how now the button component looks like (with different `style` props)

what1s1ove (changes requested):

apps/frontend/src/libs/components/button/button.tsx
Vfivew: yes here is my buttons
It seemed to me like the best solution, as now we have only three button styles instead of two colors and three styles (I would have had to add the third style anyway). We can always bring back the color prop if we need it. Or should we bring it back now?

what1s1ove: I like the idea, but how we need to provide color now? only through the classname prop?

Vfivew: Currently, the color is determined solely by the style. Before my fix, we were passing the color and then overriding it with the style (I'll add a screenshot). In my opinion, it should be the style first, and if we need to define a new color, we can add a new parameter. Currently, using three styles covers all our needs. 
plain - border/background transparent
primary - we determine color
secondary - with border but background transparent
refactor

apps/frontend/src/pages/friends/libs/components/friend/friend.tsx
```diff
@@ -55,11 +55,11 @@ const Friend: React.FC<Properties> = ({ friend }: Properties) => {
 			<div className={styles["actions"]}>
 				<Button
 					className={styles["button"]}
-					color="primary"
 					iconName={isFollowing ? "cross" : "add"}
 					label={isFollowing ? "Following" : "Follow"}
 					onClick={isFollowing ? handleUnfollow : handleFollow}
 					size="small"
+					style="primary"
```
fshabanov, line 62: primary is default in your case, no need to add it again everywhere

fshabanov (changes requested):

apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -130,7 +163,7 @@
 		top: 23px;
 		left: 33px;
 		z-index: var(--z-index-low);
-		fill: var(--main-orange);
+		color: var(--color-menu);
```
YourTeaChan, line 166: The burger button (icon) should be orange, not dark green on the mobile screen

apps/frontend/src/libs/components/button/button.tsx
what1s1ove: Okay, got it

what1s1ove (approved):

PR #193, opened 2024-02-26
fix(backend): fix error with uploading tiff and bmp format files tm-189

MappariChan:


apps/backend/src/libs/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -30,13 +30,13 @@ const fileUpload = fastifyPlugin<Options>(
 			}
 
 			const data = await request.file();
+			const buffer = await data?.toBuffer();
 
 			if (
 				data &&
+				buffer &&
 				allowedExtensions.includes(data.mimetype as ValueOf<typeof ContentType>)
```
Oleksandra-Masiuk, line 38: Please, simplify this if clause by putting this condition into a separate variable

Oleksandra-Masiuk (changes requested):

Oleksandra-Masiuk (approved):

GEOFARL (approved):

apps/backend/src/libs/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -30,13 +30,16 @@ const fileUpload = fastifyPlugin<Options>(
...
 				data &&
-				allowedExtensions.includes(data.mimetype as ValueOf<typeof ContentType>)
-			) {
-				const buffer = await data.toBuffer();
+				buffer &&
+				allowedExtensions.includes(
+					data.mimetype as ValueOf<typeof ContentType>,
+				);
 
+			if (isFileCorrect) {
```
fshabanov, line 42:
```suggestion
			if (!isFileCorrect) {
			  throw ...
```
To avoid nested if statements

what1s1ove (approved):

PR #214, opened 2024-02-27
fix(frontend): add friends screen tm-94

GEOFARL:
Make the users, who have just been followed by the current user, to disappear after switching to another tab
Make the users, whom the current user has just unfollowed, to stay on the tab until switching to another tab


apps/frontend/src/pages/friends/friends.tsx
```diff
@@ -30,7 +30,7 @@ const Friends: React.FC = () => {
 		void dispatch(actions.getFollowings());
 		void dispatch(actions.getFollowers());
 		void dispatch(actions.getPotentialFriends());
-	}, [dispatch]);
+	}, [dispatch, pathname]);
```
Oleksandra-Masiuk, line 33: The `pathname` is not used in the `useEffect`.
Why do we have to reload `Followings`, `Followers`, and `PotentialFriends` each time the `pathname` changes?
Why not reload only the type of friends that corresponds to the current value of pathname when it changes?

what1s1ove: agreed. we need to do an API call on each page separately

what1s1ove (changes requested):

Ann-Maksay (approved):

ShadowOfTheSpace (approved):

what1s1ove (approved):

PR #244, opened 2024-02-28
feat(frontend): add lightning css for css bundling tm-210 

Vfivew:
The package.json has been updated! But I checked, and everything seems to be okay.
I didn't modify some media queries because the design breaks to some extent in larger or smaller sizes.

### DIFF
--- a/apps/frontend/vite.config.ts
+++ b/apps/frontend/vite.config.ts
@@ -1,4 +1,6 @@
 import reactPlugin from "@vitejs/plugin-react";
+import browserslist from "browserslist";
+import { Features, browserslistToTargets } from "lightningcss";
 import { fileURLToPath } from "node:url";
 import { type ConfigEnv, defineConfig, loadEnv } from "vite";
 import { VitePWA } from "vite-plugin-pwa";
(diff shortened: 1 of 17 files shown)


ShadowOfTheSpace (approved):

YourTeaChan (approved):

apps/frontend/vite.config.ts
```diff
@@ -68,8 +70,21 @@ const config = ({ mode }: ConfigEnv): ReturnType<typeof defineConfig> => {
...
 	return defineConfig({
 		build: {
+			cssMinify: "lightningcss",
 			outDir: "build",
 		},
+		css: {
+			lightningcss: {
+				include:
+					Features.MediaQueries |
+					Features.Nesting |
```
what1s1ove, line 80:
```suggestion
```
provide only feature that you really need

what1s1ove (changes requested):

apps/frontend/vite.config.ts
Vfivew: I left only Media Query.
When enabled, it will also include:
1.CustomMediaQueries
2.MediaRangeSyntax
3.MediaIntervalSyntax
I think this is the minimum required pack

fshabanov (approved):

what1s1ove (approved):

PR #256, opened 2024-02-29
fix(frontend): fix add lightning css for css bundling tm-210 

Vfivew:
Fix problems with production.
For now, the only way out of the situation is.


apps/frontend/vite.config.ts
```diff
@@ -75,6 +75,9 @@ const config = ({ mode }: ConfigEnv): ReturnType<typeof defineConfig> => {
 		},
 		css: {
 			lightningcss: {
+				drafts: {
+					customMedia: true,
+				},
```
fshabanov, line 80: Maybe the issue in a deployed environment is related to the fact that drafts property was not added?
> Lightning CSS can also be configured to compile several draft specs that are not yet available natively in any browser. Because these are drafts and the syntax can still change, they must be enabled manually in your project.

[Source](https://lightningcss.dev/transpilation.html#draft-syntax:~:text=Neue%3B%0A%7D-,Draft%20syntax,-Lightning%20CSS%20can)

Vfivew: Yes, but if I remove the direct import from media.css and try to import media.css into styles.css, the draft flag will remove @custom media from styles.css.
This is the reason why I can't access custom media queries in any way wit other than directly importing from media.css.

When customMedia: false, I can use @custom media from our global styles.css, but after the build, media queries won't work.

When customMedia: true - I can't use @custom media from our global styles.css.
But i can import from media.css and build will work right

what1s1ove: It's okay to use it

apps/frontend/src/libs/components/auth-wrapper/styles.module.css
```diff
@@ -1,3 +1,5 @@
+@import url("~/assets/css/media.css");
```
what1s1ove, line 1: can we avoid this import or we really need to provide it everywhere?

what1s1ove (changes requested):

apps/frontend/src/libs/components/auth-wrapper/styles.module.css
Vfivew: The short answer is, for now, yes.
The detailed answer is here:
https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/256#discussion_r1507678698
I can't be sure that I did everything to avoid this import. But I'm currently trying to find an answer on stackoverflow.
P.S. I also created a question there.

what1s1ove: This is a new technology, not sure that we can find something on stackoverflow

Vfivew: In fact, the more I study this problem, the more it seems to me that the problem is not in lightningcss, but in the way White uses it out of the box.
I also created a discussion on Github, on their page. Perhaps there will be an answer.
From the answers that I read there, it seems to me that my theory may be correct.

what1s1ove: hey @Vfivew . any updates no this one?

Vfivew: I'm working on it right now. However, I still tend to think that the issue lies in the interaction of Vite and LightningCSS. 
P.S. I've come up with another idea of where the problem might be. Working on it!

Vfivew: Can we push this fix to main now? 
The @media isn't working and it's causing a lot of inconvenience for colleagues.
Im couldn't get rid of the import... 
The only issue is that Vite with customMedia: true simply erases my media queries when importing them into index.tsx. The build itself correctly transforms everything. 
P.S. I've also created a question directly on the Lightning CSS GitHub page, but there isn't much activity there, so I don't expect a response. 
Perhaps I can ignore the customMedia flag for development. However, it seems like not the best solution either.

PR #258, opened 2024-02-29
feat(backend/shared): drop description field from course sectionstm-251 

Vfivew:


apps/backend/src/db/migrations/20240226075220_add_course_sections_table.ts
fshabanov, line 1: You can't update existing migrations

fshabanov (changes requested):

apps/backend/src/db/migrations/20240229181005_delete_description_from_course_sections.ts
```diff
@@ -0,0 +1,23 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	COURSE_SECTIONS: "course_sections",
+} as const;
```
fshabanov, line 5:
```suggestion
const TABLE_NAME = "course_sections";
```
Same for ColumnName

fshabanov (changes requested):

apps/backend/src/db/migrations/20240229181005_delete_description_from_course_sections.ts
Vfivew: I fixed it, but now I still have fix linter.
P.S. Probably, the linter breaks down when you define yourself as a performer after creating a PR, and not before. My mistake...

GEOFARL (approved):

PR #261, opened 2024-02-29
fix(frontend/backend): fix profile screen tm-42

Sytnikovakatya:
- Solve the issue when trying to update the profile in the deployed app
- Fix style of image on the header and the profile picture on the "Profile" page with rectangular photos


Vfivew (approved):

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -134,13 +134,13 @@ class UserController extends BaseController {
 	private async updateUser(
 		options: APIHandlerOptions<{
 			body: UserProfileRequestDto;
-			params: { id: string };
+			params: {
+				id: number;
+			};
```
fshabanov, line 139: Param is always string, since it is a part of URL

fshabanov (changes requested):

apps/frontend/src/modules/users/users-api.ts
```diff
@@ -40,7 +40,7 @@ class UserApi extends BaseHTTPApi {
 		payload: UserProfileRequestDto,
 	): Promise<UserAuthResponseDto> {
 		const response = await this.load(
-			this.getFullEndpoint(`${UsersApiPath.ROOT}/${id}`, {}),
+			this.getFullEndpoint(UsersApiPath.$ID, { id: String(id) }),
```
ShadowOfTheSpace, line 43:
```suggestion
			this.getFullEndpoint(`${UsersApiPath.$ID}${id}`, { }),
```

ShadowOfTheSpace (approved):

apps/frontend/src/modules/users/users-api.ts
what1s1ove: why? it's better to use `getFullEndpoint` instead

what1s1ove (approved):

PR #268, opened 2024-03-01
feat(backend): Convert text to varchar in DB for fields with a maximum length tm-265

GEOFARL:


apps/backend/src/db/migrations/20240301085056_convert_text_to_string.ts
```diff
@@ -0,0 +1,40 @@
+import { type Knex } from "knex";
+
+import { UserValidationRule } from "~/modules/users/users.js";
```
what1s1ove, line 3:
```suggestion
```
you can not use external things in the migration

what1s1ove (changes requested):

apps/backend/src/db/migrations/20240301085056_convert_text_to_string.ts
```diff
@@ -0,0 +1,39 @@
+import { type Knex } from "knex";
+
+const TableName = "user_details";
```
tarmiel, line 3:
```suggestion
const TABLE_NAME= "user_details";
```

tarmiel (changes requested):

what1s1ove (approved):

PR #284, opened 2024-03-02
fix(frontend/backend): Add Friend page tm-40

GEOFARL:
- Updated the "Courses" title to be 20px
- Made the "Back" button to be as on the "Chat" and "Course details" pages and change after hovering
- Made the Follow" and "Following" buttons not to change regarding the user's name length (size should be 120px 32px)
- Made the GET /user-courses/{userId} response body not to be empty even when user has courses


apps/frontend/src/pages/user/styles.module.css
```diff
@@ -47,15 +47,18 @@
...
 .return-button {
 	align-self: flex-start;
-	padding: 0;
+	width: 25px;
+	height: 25px;
 	margin-bottom: 17px;
 	color: var(--color-text-200);
-	background-color: transparent;
-	border-color: var(--color-background);
+	background-color: #e0e0e0;
```
Vfivew, line 54: Do we need new varieable for it maybe?

Sytnikovakatya (approved):

apps/frontend/src/pages/user/styles.module.css
```diff
@@ -47,15 +47,18 @@
...
-.return-button:hover {
-	color: var(--color-text-200);
-	background-color: transparent;
-	border-color: var(--color-background);
+.courses-title {
+	font-size: var(--font-size-400);
+}
+
+.follow-button {
+	width: 120px;
```
what1s1ove, line 63:
```suggestion
	min-width: 120px;
```
never add a fix width

what1s1ove (changes requested):

apps/frontend/src/pages/user/styles.module.css
GEOFARL: I do not think that it is necessary

KrasKa2: @GEOFARL  Why do you think so? We have a rule, that everything should be the same. So if we use vars for color, we should use it everywhere

what1s1ove (approved):

PR #292, opened 2024-03-03
feat(backend): add edx vendor tm-211

Ann-Maksay:
At the moment this is a draft service that is responsible for requests to edX API made based on [swagger documentation](https://courses.edx.org/api-docs/)
And it does not cover all necessary functionality.
Available:
1. Get a list of courses
2. Get course information by ID.

Not Available:
1. Search by course
2. Get course sections.

Displaying the list of courses:
show_courses

Course details:
deteils

To make it clearer, the data I have on the course. The route with blocks returns an authorisation error, so it is not possible to get the course sections.
image


apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,124 @@
...
+		return mappedItem;
+	}
+
+	private mapToCourse(item: Record<string, unknown>): Course {
+		const course = this.mapItem(item, EdxCourseFieldsMapping);
+
+		const media = item["media"] as Record<string, unknown>;
+		const courseImage = media["course_image"] as Record<string, unknown>;
+
+		course.image = `${this.baseUrl}${courseImage["uri"] as string}`;
```
Ann-Maksay, line 88: Actually I did it very roughly, as I think, because the object received from the real API is different and the mapping function will be deeply refactored.

apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,124 @@
...
+			`${this.baseUrl}${EdxApiPath.COURSES}${id}`,
+			{},
+		);
+		const item = (await result.json()) as Record<string, unknown>;
+
+		return this.mapToCourse(item);
+	}
+
+	public async getCourseSections(vendorId: string): Promise<CourseSection[]> {
+		await Promise.resolve(vendorId);
```
Ann-Maksay, line 105: All this is necessary to avoid errors related to types that methods return in VendorService.

apps/backend/src/db/migrations/20240302150600_add_edx_vendor.ts
```diff
@@ -0,0 +1,18 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	VENDORS: "vendors",
+};
```
Vfivew, line 5:
```suggestion
const TABLE_NAME = "vendors"
```

Sytnikovakatya (approved):

Vfivew (approved):

apps/backend/src/modules/vendors/libs/enums/edx-api-path.enum.ts
```diff
@@ -0,0 +1,6 @@
+const EdxApiPath = {
+	COURSES: "api/courses/v1/courses/",
+	COURSES_SECTIONS: "api/courses/v1/blocks/",
```
fshabanov, line 3: `api/courses/v1` should be a part of URL in ENV

fshabanov (changes requested):

apps/backend/src/modules/vendors/libs/enums/edx-api-path.enum.ts
Ann-Maksay: The URL from .env is also used to create the image path, but there is no part with `api/courses/v1/`. This is the main reason why I split the route this way. I really should change it @fshabanov ?
Image route: `${this.baseUrl}${courseImage["uri"]`

GEOFARL (approved):

apps/backend/src/db/migrations/20240302150600_add_edx_vendor.ts
```diff
@@ -0,0 +1,16 @@
+import { type Knex } from "knex";
+
+const TableName = "vendors";
```
what1s1ove, line 3: it is not a enum

apps/backend/src/db/migrations/20240302150600_add_edx_vendor.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const KEY_COLUMN = "key";
+const row = { key: "edx", name: "edX", url: "https://www.edx.org" };
```
what1s1ove, line 6: should be a constant

apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,124 @@
...
+
+	private getHeaders(): Headers {
+		const headers = new Headers();
+
+		headers.append(HTTPHeader.CONTENT_TYPE, ContentType.JSON);
+
+		return headers;
+	}
+
+	private load(url: string, query: Record<string, unknown>): Promise<Response> {
```
what1s1ove, line 42: use real types everywhere in this file. describe fields (at least that we use)

apps/backend/src/modules/vendors/libs/enums/edx-course-field.enum.ts
```diff
@@ -0,0 +1,7 @@
+const EdxCourseField: Record<string, string> = {
```
what1s1ove, line 1:
```suggestion
const EdxCourseField = {
```

what1s1ove (changes requested):

apps/backend/src/modules/vendors/edx.service.ts
Ann-Maksay: I'm not sure how I can best describe the type for query. So far I've got something like this:
```
type EdxCourseQuery = {
  search_term?: string;
  page?: number;
  pageSize?: number;
};
```
But as they say, if I want to make an optional key, it's better to ask how not to make it

apps/backend/src/db/migrations/20240302150600_add_edx_vendor.ts
```diff
@@ -0,0 +1,17 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "vendors";
+
+const KEY_COLUMN = "key";
+
+const Row = { key: "edx", name: "edX", url: "https://www.edx.org" } as const;
```
what1s1ove, line 7: it is not an enum. it's a constant

apps/backend/src/modules/vendors/libs/types/edx-course-response-dto.type.ts
```diff
@@ -0,0 +1,19 @@
+type EdxImage = {
+	[key: string]: unknown;
```
what1s1ove, line 2:
```suggestion
```
do not use an index signature. describe only keys that we need

apps/backend/src/modules/vendors/libs/types/edx-course-response-dto.type.ts
```diff
@@ -0,0 +1,19 @@
...
+	uri: string;
+};
+
+type EdxMedia = {
```
what1s1ove, line 6: each type should be in a separate file

apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,119 @@
...
+			payload: null,
+			query,
+		});
+	}
+
+	private async loadResults(
+		url: string,
+		query: Record<string, unknown>,
+	): Promise<EdxCourseResponseDto[]> {
+		const result = await this.load(url, query);
```
what1s1ove, line 51: make `load` as a generic. Check the http service

apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,119 @@
...
+	private load(url: string, query: Record<string, unknown>): Promise<Response> {
+		return this.http.load(url, {
+			headers: this.getHeaders(),
+			method: "GET",
+			payload: null,
+			query,
+		});
+	}
+
+	private async loadResults(
```
what1s1ove, line 47: why do we need a separate method for it if we use it only in `getCourses`

apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,119 @@
...
+		url: string,
+		query: Record<string, unknown>,
```
what1s1ove, line 49: use a real type

apps/backend/src/modules/vendors/edx.service.ts
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	public async getCourseSections(vendorId: string): Promise<CourseSection[]> {
+		await Promise.resolve(vendorId);
+
+		return [];
+	}
+
+	public async getCourses(search: string): Promise<Course[]> {
+		const query: Record<string, unknown> = {
```
what1s1ove, line 106: use a real type

apps/backend/src/modules/vendors/edx.service.ts
what1s1ove: you can use an optional keys here

what1s1ove (changes requested):

apps/backend/src/modules/vendors/libs/mapper/edx-course-fields.mapper.ts
```diff
@@ -0,0 +1,15 @@
+import { EdxCourseField } from "~/modules/vendors/libs/enums/enums.js";
+import {
+	type CourseFieldsMapping,
+	type EdxCourseResponseDto,
+} from "~/modules/vendors/libs/types/types.js";
+
+const EdxCourseFieldsMapping: CourseFieldsMapping = {
```
what1s1ove, line 7: should be a map, not an enum

what1s1ove: https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/320#discussion_r1513019239

apps/backend/src/modules/vendors/edx.service.ts
what1s1ove: return Promise.resolve([]) ?

what1s1ove (changes requested):

apps/backend/src/modules/vendors/edx.service.ts
Ann-Maksay: I tried it first, but linter won't let it pass. This method must have a string argument, must have an asynchronous operation and must return an array.

apps/backend/src/modules/vendors/libs/mapper/edx-course-fields.mapper.ts
```diff
@@ -0,0 +1,15 @@
+import { EdxCourseField } from "~/modules/vendors/libs/enums/enums.js";
+import {
+	type CourseFieldsMapping,
+	type EdxCourseResponseDto,
+} from "~/modules/vendors/libs/types/types.js";
+
+const EdxCourseFieldsMap: CourseFieldsMapping = {
```
Vfivew, line 7: Vlad taught me, and now I can pass on this knowledge to you!
https://dev.to/what1s1ove/javascript-naming-conventions-are-important-45n
```suggestion
const edxCourseToCourse: CourseFieldsMapping = {
```

Vfivew (changes requested):

apps/backend/src/modules/vendors/libs/mapper/edx-course-fields.mapper.ts
```diff
@@ -0,0 +1,15 @@
+import { EdxCourseField } from "~/modules/vendors/libs/enums/enums.js";
```
what1s1ove, line 1: should be in `maps` folder and with `map` prefix

apps/backend/src/modules/vendors/edx.service.ts
what1s1ove: you can remove this argument at all

what1s1ove (changes requested):

KrasKa2 (approved):

what1s1ove (approved):

PR #316, opened 2024-03-04
feat: add app domain and ssl tm-287

fshabanov:


.ebextensions/alb-http-to-https-redirection.config
what1s1ove, line 1: add labler to this folder

what1s1ove (approved):

PR #352, opened 2024-03-05
feat(frontend): change button icon on hover tm-319

KrasKa2:
task-319


Vfivew (approved):

Sytnikovakatya (approved):

Oleksandra-Masiuk:
Please, check out my answer in the chat. Maybe it will be helpful

ShadowOfTheSpace (commented):
Can you use here className prop from the Icon component? I think you can create a new prop for the button (iconClassName) and pass it to the Icon component. I think it can be helpful for styling Icon outside of the Button component.

P.S. I saw answer in chat and I think my comment is not helpful for this situation(

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -58,7 +61,8 @@ const Button: React.FC<Properties> = ({
 					onClick={onClick}
 					type={type}
 				>
-					{icon}
+					<div className={styles["icon"]}>{icon}</div>
+					<div className={styles["hoverIcon"]}>{hoverIcon}</div>
```
what1s1ove, line 65: we will not add an additional icon here. find a better solution

you can add the classname prop to the Icon component to try to handle this issue

what1s1ove (changes requested):

KrasKa2:
> Can you use here className prop from the Icon component? I think you can create a new prop for the button (iconClassName) and pass it to the Icon component. I think it can be helpful for styling Icon outside of the Button component.
> 
> P.S. I saw answer in chat and I think my comment is not helpful for this situation, but if you change icon as Oleksandra said you can style it via iconClassName

@ShadowOfTheSpace your advice works only with Oleksandara's advice))

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -39,7 +41,9 @@ const Button: React.FC<Properties> = ({
 		isDisabled && styles["disabled"],
 	);
 
-	const icon = iconName ? <Icon name={iconName} /> : null;
+	const icon = iconName ? (
+		<Icon className={iconClassName || ""} name={iconName} />
```
fshabanov, line 45:
```suggestion
		<Icon className={iconClassName ?? ""} name={iconName} />
```

fshabanov (changes requested):

apps/frontend/src/libs/components/button/button.tsx
what1s1ove: change props in Icon components

what1s1ove (changes requested):

apps/frontend/src/libs/components/button/button.tsx
KrasKa2: > change props in Icon components
@what1s1ove  I have done changes suggested by @fshabanov and I don't understand what alse I have to do

Oleksandra-Masiuk (approved):

what1s1ove (approved):

PR #356, opened 2024-03-06
feat(backend/frontend/shared): real time chat functionality tm-113

ShadowOfTheSpace:
- Case with 4 different connections from 4 users
4 chats

- Case with 3 different connections from 2 users
2 chats 3 connections


apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -13,6 +13,7 @@ import { type Config } from "~/libs/modules/config/config.js";
 import { type Database } from "~/libs/modules/database/database.js";
 import { HTTPCode, HTTPError } from "~/libs/modules/http/http.js";
 import { type Logger } from "~/libs/modules/logger/logger.js";
+import { socketService } from "~/libs/modules/socket/socket.js";
```
what1s1ove, line 16: inject in through the constructor (the same as user-service, as you can see)

apps/backend/src/libs/modules/socket/socket-service.module.ts
```diff
@@ -0,0 +1,42 @@
+import { type Server as HttpServer } from "node:http";
+import { type Socket, Server as SocketServer } from "socket.io";
+
+import { SocketEvent, SocketNamespace } from "~/libs/modules/socket/socket.js";
+import { type ValueOf } from "~/libs/types/types.js";
+
+class SocketService {
+	private chatMessageHandler = (socket: Socket): void => {
```
what1s1ove, line 8: why the arrow function?

apps/frontend/src/libs/modules/store/middlewares/chat-socket.middleware.ts
```diff
@@ -0,0 +1,51 @@
+import {
+	type Middleware,
+	type PayloadAction,
+	isAction,
+} from "@reduxjs/toolkit";
+
+import {
+	SocketEvent,
+	SocketNamespace,
+	socket,
```
what1s1ove, line 10:
```suggestion
```
services should be injected aiv middleware parameters

apps/frontend/src/libs/modules/store/middlewares/chat-socket.middleware.ts
```diff
@@ -0,0 +1,51 @@
...
+} from "~/libs/modules/socket/socket.js";
+import { type AppDispatch } from "~/libs/types/types.js";
+import { type ChatMessageItemResponseDto } from "~/modules/chat-messages/chat-messages.js";
+import { actions as chatsActions } from "~/modules/chats/chats.js";
+
+const chatSocketInstance = socket.getInstance(SocketNamespace.CHAT);
+
+type SocketMiddlewareParameters = {
```
what1s1ove, line 18:
```suggestion
type Parameters = {
```

apps/frontend/src/modules/chats/slices/actions.ts
```diff
@@ -47,10 +46,102 @@ const createChat = createAsyncThunk<
...
+			lastMessage: newMessage,
+			unreadMessageCount:
+				chat.interlocutor.id === newMessage.senderUser.id
+					? Number(chat.unreadMessageCount) + ONE_UNREAD_MESSAGE
+					: chat.unreadMessageCount,
+		};
+	});
+
+	updatedChats.sort(
+		({ lastMessage: lastMessage1 }, { lastMessage: lastMessage2 }) => {
```
what1s1ove, line 88:
```suggestion
		(a, b) => {
```

what1s1ove (changes requested):

apps/backend/src/libs/modules/socket/socket-service.module.ts
ShadowOfTheSpace: Because of eslint error (this function does not access 'this' prop of class, and eslint said that it need to be an arrow function if it not access 'this'. But maybe I need to change it to static function?

Vfivew (approved):

PR #365, opened 2024-03-06
fix(frontend): The element on the Friend profile section move regarding the user name tm-327

GEOFARL:
image
image
image


GEOFARL:
Need to clarify some moments with QAs

Sytnikovakatya (approved):

what1s1ove (changes requested):
handle it with https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

GEOFARL:
> handle it with https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

contacted qa team about it before applying absolute positioning, they said that it is not acceptable

ShadowOfTheSpace (approved):

Vfivew (approved):

what1s1ove (approved):

PR #366, opened 2024-03-06
feat(backend/frontend/shared): add search field to chats page tm-294

Sytnikovakatya:
- Add the "Search" field to the "Chats" page
- Add the possibility to search chats by user's name

Screenshot_11


apps/backend/src/modules/chats/chat.repository.ts
```diff
@@ -126,9 +132,41 @@ class ChatRepository implements Repository<ChatEntity> {
...
+								.orWhere(
+									"firstUser:userDetails.lastName",
+									"ilike",
+									`%${search}%`,
+								);
+						})
+						.andWhere({ secondUserId: userId })
+						.orWhere((builder) => {
+							void builder
+								.where(
```
fshabanov, line 153:
```suggestion
								.whereILike(
```
And I think we can do this without `if ... else` statement, because if `search` is empty, then `.whereILike("column", "%%")` will return everything

apps/frontend/src/pages/chats/libs/components/chat-search-bar/chat-search-bar.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const ChatSearchBar: React.FC = () => {
+	const dispatch = useAppDispatch();
+
+	const { control, errors, handleSubmit } = useAppForm({
+		defaultValues: DEFAULT_SEARCH_CHAT_PAYLOAD,
+		mode: "onChange",
+	});
+
+	const handleSearchCourses = (filterFormData: ChatSearchDto): void => {
```
fshabanov, line 27:
```suggestion
	const handleSearchChats = (filterFormData: ChatSearchDto): void => {
```
We are searching for chats here

fshabanov (changes requested):

apps/backend/src/modules/chats/chat.repository.ts
Sytnikovakatya: Yes, it works without ``if ... else``. Thanks

apps/frontend/src/modules/chats/slices/actions.ts
```diff
@@ -9,19 +9,22 @@ import { type ChatMessageItemResponseDto } from "~/modules/chat-messages/chat-me
...
+	type ChatSearchDto,
 } from "../libs/types/types.js";
 import { name as sliceName } from "./chats.slice.js";
 
 const ONE_UNREAD_MESSAGE = 1;
 
 const getAllChats = createAsyncThunk<
 	{ items: ChatGetAllItemResponseDto[] },
-	undefined,
+	ChatSearchDto | undefined,
```
what1s1ove, line 20:
```suggestion
	ChatSearchDto,
```

apps/frontend/src/pages/chats/libs/components/chat-search-bar/chat-search-bar.tsx
```diff
@@ -0,0 +1,61 @@
...
+import {
+	type ChatSearchDto,
+	DEFAULT_SEARCH_CHAT_PAYLOAD,
+	SEARCH_CHAT_DELAY_MS,
+	actions as chatsActions,
+} from "~/modules/chats/chats.js";
+
+import styles from "./styles.module.css";
+
+const handleFormSubmit = (event_: FormEvent<HTMLFormElement>): void => {
```
what1s1ove, line 15: should be inside the component

packages/shared/src/modules/chats/libs/types/chat-search-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+type ChatSearchDto = {
```
what1s1ove, line 1: add request or response. always specify a type of DTO

what1s1ove (changes requested):

apps/backend/src/modules/chats/chat.repository.ts
ShadowOfTheSpace: @fshabanov I agree that query `where field_name ilike '%%'` (with empty search template) will give the same result as query without that `where` statement. But I checked it in the execution plan and it is different from query without `where` statement. It is really filtering rows by excluding null values (`where field_name ilike '%%'` is equivalent to `where field_name is not null`. So I think query with empty `'%%'` is less efficient than query without ilike at all. In time it is 1.5 - 2 times slower than query without filtering. I checked it on a simple query with small row count, but I think that on chat querying (as it is on our situation) it will be more notable. Is it worth it?

what1s1ove (approved):

PR #379, opened 2024-03-06
feat(backend/frontend/shared): add sex property to the user tm-343

MappariChan:


apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -42,6 +42,9 @@ import { AuthApiPath } from "./libs/enums/enums.js";
  *          nickname:
  *            type: string
  *            nullable: true
+ *          sex:
+ *            type: string
+ *            enum: [Male, Female, Prefere not to say]
```
fshabanov, line 47:
```suggestion
 *            enum: [Male, Female, Prefer not to say]
```

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -122,6 +122,9 @@ class UserController extends BaseController {
 	 *                  type: string
 	 *                nickname:
 	 *                  type: string
+	 *                sex:
+	 *                  type: string
+	 *                  enum: [Male, Female, Prefere not to say]
```
fshabanov, line 127:
```suggestion
	 *                  enum: [Male, Female, Prefer not to say]
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,68 @@
...
+		styles["heading"],
+		hasVisuallyHiddenLabel && "visually-hidden",
+	);
+
+	const deffaulValue = field.value ?? "";
+
+	return (
+		<label className={styles["container"]}>
+			<span className={labelClasses}>{label}</span>
+			<select className={selectClasses} {...field} value={deffaulValue}>
```
fshabanov, line 57:
```suggestion
			<select className={selectClasses} {...field} value={defaultValue}>
```

packages/shared/src/modules/users/libs/validation-schemas/user-profile.validation-schema.ts
```diff
@@ -11,6 +15,7 @@ type UserProfileRequestValidationDto = {
 			z.ZodUnion<[z.ZodString, z.ZodLiteral<null>]>
 		>
 	>;
+	sex: z.ZodNullable<z.ZodEnum<["Male", "Female", "Prefer not to say"]>>;
```
fshabanov, line 18:
```suggestion
	sex: z.ZodNullable<z.ZodEnum<ValueOf<typeof UserSex>>>;
```

fshabanov (changes requested):

apps/backend/src/db/migrations/20240306140243_add_sex_to_the_user_details_table.ts
```diff
@@ -0,0 +1,31 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "user_details";
+
+const COLUMN_NAME = "sex";
+
+const UserSex = {
+	FEMALE: "Female",
```
what1s1ove, line 8: should be kebab case

apps/backend/src/db/migrations/20240306140243_add_sex_to_the_user_details_table.ts
```diff
@@ -0,0 +1,31 @@
...
+	MALE: "Male",
+	PREFER_NOT_TO_SAY: "Prefer not to say",
+} as const;
+
+async function up(knex: Knex): Promise<void> {
+	await knex.schema.alterTable(TABLE_NAME, (table) => {
+		table
+			.enum(COLUMN_NAME, [
```
what1s1ove, line 16: Object.values

packages/shared/src/modules/users/libs/enums/user-sex.enum.ts
```diff
@@ -0,0 +1,7 @@
+const UserSex = {
+	FEMALE: "Female",
```
what1s1ove, line 2: should be kebab case

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -138,6 +140,20 @@ const Profile: React.FC = () => {
...
 							/>
+							<Select
+								color="light"
+								control={control}
+								errors={errors}
+								label="Sex"
+								name="sex"
+								placeholder="Select sex"
+							>
+								<option value={UserSex.MALE}>{UserSex.MALE}</option>
```
what1s1ove, line 151: Object values for options

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,68 @@
...
+				<option disabled hidden value="">
```
what1s1ove, line 58: we need to add it only in case if we have `isRequired` = true

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,68 @@
...
+	hasVisuallyHiddenLabel?: boolean;
+	label: string;
+	name: FieldPath<T>;
+	placeholder?: string;
+};
+
+const Select = <T extends FieldValues>({
+	children,
+	className,
+	color = "light",
```
what1s1ove, line 28: for what need this props?

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,68 @@
...
+	type FieldValues,
+} from "react-hook-form";
+
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+import { useFormController } from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+type Properties<T extends FieldValues> = {
+	children: React.ReactNode;
```
what1s1ove, line 14:
```suggestion
```

what1s1ove (changes requested):

Sytnikovakatya (approved):

ShadowOfTheSpace (approved):

Vfivew (approved):

packages/shared/src/modules/users/libs/validation-schemas/user-profile.validation-schema.ts
ShadowOfTheSpace: Use ZodNativeEnum instead

apps/frontend/src/libs/components/select/select.tsx
MappariChan: do you mean to add the default option only when `isRequired = true`?

apps/frontend/src/libs/components/select/libs/types/option.type.ts
```diff
@@ -0,0 +1,6 @@
+type Option = {
```
what1s1ove, line 1:
```suggestion
type SelectOption = {
```
should be in `src/libs/types`

apps/frontend/src/libs/components/select/select.tsx
what1s1ove: right

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -83,6 +93,15 @@ const Profile: React.FC = () => {
 
 	useAppTitle(AppTitle.PROFILE);
 
+	const sexSelectOptions = useMemo<Option[]>(() => {
+		return Object.values(UserSex).map((sex) => {
+			return {
+				label: convertKebabCaseToRegularString(sex, { capitalize: true }),
```
what1s1ove, line 99: you do not need this helper. you need to use a map datastructure

packages/shared/src/libs/helpers/convert-kebab-case-to-regular-string/convert-kebab-case-to-regular-string.helper.ts
```diff
@@ -0,0 +1,26 @@
+import { Separator, StringConstant } from "./libs/enums/enums.js";
+import { type Options } from "./libs/types/types.js";
+
+const convertKebabCaseToRegularString = (
```
what1s1ove, line 4:
```suggestion
```

what1s1ove (changes requested):

packages/shared/src/modules/users/libs/validation-schemas/user-profile.validation-schema.ts
```diff
@@ -95,6 +100,7 @@ const userProfile = z.object<UserProfileRequestValidationDto>({
 				.or(z.literal(null)),
 		)
 		.nullable(),
+	sex: z.nativeEnum(UserSex).nullable(),
```
fshabanov, line 103: Let's use `enum` as we do in other places

fshabanov (changes requested):

packages/shared/src/modules/users/libs/validation-schemas/user-profile.validation-schema.ts
MappariChan: I can't do as you said before
image
There is another way how to do this and it's something like that
image
And for declaring schema types we didn't use ZodEnum before
image
So can i leave NativeEnum because it's more readable?

KrasKa2 (approved):

GEOFARL (approved):

apps/frontend/src/libs/components/select/styles.module.css
```diff
@@ -0,0 +1,57 @@
...
+	padding: 11px 12px;
+	font-family: inherit;
+	font-weight: 500;
+	line-height: 1.1;
+	color: var(--color-text-200);
+	appearance: none;
+	cursor: pointer;
+	background-color: var(--color-background);
+	background-image: none;
+	background-image: url("~/assets/icons/select-arrow.svg");
```
fshabanov, line 26: Do we need to have both?

packages/shared/src/modules/users/libs/validation-schemas/user-profile.validation-schema.ts
fshabanov: I think we already have [type, ...type[]] somewhere else, so it is okay to use it

fshabanov (changes requested):

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -83,6 +87,21 @@ const Profile: React.FC = () => {
 
 	useAppTitle(AppTitle.PROFILE);
 
+	const sexSelectOptions = useMemo<SelectOption[]>(() => {
+		const userSexToTheSelectLabel: Record<ValueOf<typeof UserSex>, string> = {
```
what1s1ove, line 91:
```suggestion
		const userSexToTheSelectLabel = {
```
move it to libs

what1s1ove, line 91:
```suggestion
		const userSexToReadable: Record<ValueOf<typeof UserSex>, string> = {
```

what1s1ove (changes requested):

what1s1ove (approved):

PR #387, opened 2024-03-06
feat(frontend/backend/shared): friends activity feed comment tm-308

tarmiel:
demo


Sytnikovakatya (approved):

apps/backend/src/libs/types/service.type.ts
```diff
@@ -1,6 +1,6 @@
 type Service<T = unknown> = {
 	create(payload: unknown): Promise<T>;
-	delete(id: number): Promise<boolean>;
+	delete(id: number, payload: unknown): Promise<boolean>;
```
what1s1ove, line 3:
```suggestion
	delete(id: number): Promise<boolean>;
```
you can not change this implement type

apps/frontend/src/pages/feed/libs/components/activity-comment-form/activity-comment-form.tsx
```diff
@@ -0,0 +1,58 @@
...
+	isLoading: boolean;
+	onSubmit: (payload: Pick<CommentCreateRequestDto, "text">) => void;
+};
+
+const ActivityCommentForm: React.FC<Properties> = ({
+	isLoading,
+	onSubmit,
+}: Properties) => {
+	const { control, errors, handleSubmit, reset } = useAppForm({
+		defaultValues: {
```
what1s1ove, line 18: move to constants

packages/shared/src/modules/activities/libs/types/activity-counts.type.ts
```diff
@@ -0,0 +1,5 @@
+type ActivityCounts = {
```
what1s1ove, line 1: activity comment count ?

what1s1ove (changes requested):

apps/backend/src/libs/types/service.type.ts
tarmiel: For delete in service we need for example current user info to check whether user is allowed to delete only his comments

packages/shared/src/modules/activities/libs/types/activity-counts.type.ts
what1s1ove: agreed. let's combine them. but do not forget to say about it in the like PR

apps/backend/src/libs/types/service.type.ts
what1s1ove: do it with a separate call to db for now

apps/backend/src/modules/activities/libs/constants/empty-comment-count.constant.ts
```diff
@@ -0,0 +1,3 @@
+const EMPTY_COMMENT_COUNT = 0;
+
+export { EMPTY_COMMENT_COUNT };
```
Ann-Maksay, line 3: Since this can also be used for other counters, I'd like to summarize it as EMPTY_COUNT

what1s1ove (approved):

PR #397, opened 2024-03-07
feat(frontend/shared): add pagination with query tm-307

Vfivew:


packages/shared/src/libs/enums/pagination-value.enum.ts
```diff
@@ -1,6 +1,7 @@
 const PaginationValue = {
 	DEFAULT_COUNT: 10,
 	DEFAULT_PAGE: 1,
+	PAGE_NOT_EXIST: 0,
```
YourTeaChan, line 4:
```suggestion
	PAGE_NOT_EXISTS: 0,
```

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,8 +39,23 @@ const usePagination: UsePagination = ({
...
+		} else if (pagesCount !== PaginationValue.PAGE_NOT_EXISTS) {
+			history("", { replace: true });
+			setPage(PaginationValue.DEFAULT_PAGE);
+		}
+	}, [location.search, page, pagesCount]);
+
 	const handlePageChange = (page: number): void => {
 		setPage(page);
+
+		history(`?page=${page}`, { replace: true });
```
what1s1ove, line 58: we have useEffect with a page dependency, and we set it here, after the set-state function. it looks strange. you need to leave it in one place

what1s1ove (changes requested):

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/get-pages-valid.helper.ts
```diff
@@ -0,0 +1,16 @@
+import { PaginationValue } from "~/libs/enums/enums.js";
+
+const getPagesValid = ({
+	pageFromUrl,
+	searchParameters,
+}: {
+	pageFromUrl: number;
+	searchParameters: URLSearchParams;
+}): boolean => {
+	const isPageParameterExist = searchParameters.has("page");
```
fshabanov, line 10: It doesn't make any sense, you send `pageFromUrl` to this helper, which comes from URL, no need to check it again

packages/shared/src/libs/enums/pagination-value.enum.ts
```diff
@@ -1,6 +1,7 @@
 const PaginationValue = {
 	DEFAULT_COUNT: 10,
 	DEFAULT_PAGE: 1,
+	PAGE_NOT_EXISTS: 0,
```
fshabanov, line 4: Do we need it? If the page is not specified, then we will use `DEFAULT_PAGE`

fshabanov (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,8 +39,23 @@ const usePagination: UsePagination = ({
 	});
 	const pages = getPagesRange(pagesCut.start, pagesCut.end);
 
+	useEffect(() => {
+		const searchParameters = new URLSearchParams(location.search);
+		const pageFromUrl = Number(searchParameters.get("page"));
+		const isValidPage = getPagesValid({ pageFromUrl, searchParameters });
+
+		if (pageFromUrl <= pagesCount && isValidPage) {
```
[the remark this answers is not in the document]
Vfivew: Yes. I redirect to 1-th page. I will show gif a bit later, when i resolve all fix

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
Vfivew: I will add another logic where there won't be handlePageChange. Instead of buttons in pagination, I would use links.

packages/shared/src/libs/enums/pagination-value.enum.ts
Vfivew: When we have first render, i got total=0 in state .
So when we use `const pagesCount = Math.ceil(totalCount / pageSize);` - im got pagesCount=0.
I deliberately check for zero to avoid entering the 'else' block during the render.
So I need this constant. I can replace it with use-pagination.hook.ts if it is better.

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/get-pages-valid.helper.ts
```diff
@@ -0,0 +1,16 @@
+import { PaginationValue } from "~/libs/enums/enums.js";
+
+const getPagesValid = ({
```
Oleksandra-Masiuk, line 3:
```suggestion
const checkIsPageValid = ({
```

Oleksandra-Masiuk (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,8 +39,23 @@ const usePagination: UsePagination = ({
 	});
 	const pages = getPagesRange(pagesCut.start, pagesCut.end);
 
+	useEffect(() => {
+		const searchParameters = new URLSearchParams(location.search);
```
ShadowOfTheSpace, line 43: Use `useSearchParams` hook instead

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -20,7 +29,8 @@ const usePagination: UsePagination = ({
 	totalCount,
 }) => {
 	const [page, setPage] = useState<number>(PaginationValue.DEFAULT_PAGE);
-
+	const location = useLocation();
+	const history = useNavigate();
```
ShadowOfTheSpace, line 33: Rename variable `history` to `navigation`

ShadowOfTheSpace (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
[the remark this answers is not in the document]
Vfivew: Well, here is a gif.
pagin

apps/frontend/src/libs/components/pagination/libs/components/pagination-item/pagination-item.tsx
```diff
@@ -22,13 +25,9 @@ const PaginationItem: React.FC<Properties> = ({
...
 	return (
 		<li className={styles["item-container"]}>
-			<button
-				className={linkClasses}
-				disabled={isDisabled}
-				onClick={onPageChange}
-			>
-				{label}
-			</button>
+			<Link className={linkClasses} isDisabled={isDisabled ?? false} to={to}>
```
GEOFARL, line 28: maybe, it'll be better to give to `isDisabled` a default parameter

Sytnikovakatya (approved):

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -26,9 +28,10 @@ const Link: React.FC<Properties> = ({
 				isActive && activeClassName,
 				className,
 				styles["link"],
+				isDisabled && styles["disabled"],
 			);
 		},
-		[activeClassName, className],
+		[activeClassName, className, isDisabled],
```
what1s1ove, line 34: in case of disabled we need to render a span

apps/frontend/src/libs/components/link/styles.module.css
```diff
@@ -1,3 +1,7 @@
 .link {
 	text-decoration: none;
 }
+
+.disabled {
```
what1s1ove, line 5: what about the keyboard and other devices? we need to render span instead

apps/frontend/src/libs/components/pagination/libs/components/pagination-item/pagination-item.tsx
```diff
@@ -22,13 +25,9 @@ const PaginationItem: React.FC<Properties> = ({
 
 	return (
 		<li className={styles["item-container"]}>
-			<button
```
what1s1ove, line 25: it should be a button. you need to change it via handler

what1s1ove (changes requested):

apps/frontend/src/libs/components/pagination/libs/components/pagination-item/pagination-item.tsx
Vfivew: That is, all pagination items are buttons?
P.S. I just saw that links are used more often, so I changed it. This makes it easier to control what is in the query

KrasKa2 (approved):

apps/frontend/src/libs/components/pagination/libs/components/pagination-item/pagination-item.tsx
```diff
@@ -1,19 +1,22 @@
+import { type AppRoute } from "~/libs/enums/app-route.enum.js";
 import { getValidClassNames } from "~/libs/helpers/helpers.js";
+import { type ValueOf } from "~/libs/types/types.js";
 
+import { Link } from "../../../../link/link.js";
```
fshabanov, line 5: Fix the import route

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/cheak-is-vallid-page/cheak-is-vallid-page.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { PaginationValue } from "~/libs/enums/enums.js";
+
+const checkIsPageValid = (pageFromUrl: number): boolean => {
+	return pageFromUrl > PaginationValue.DEFAULT_PAGE;
```
fshabanov, line 4: It is a simple check and we use this helper only in 1 place, I don't think we need a separate helper for that

fshabanov (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,6 +38,23 @@ const usePagination: UsePagination = ({
...
+			queryParameters.size > QUERY_SECTION_SIZE
+		) {
+			setSearchParameters();
+			setPage(PaginationValue.DEFAULT_PAGE);
+		}
+	}, [page, pagesCount]);
+
 	const handlePageChange = (page: number): void => {
 		setPage(page);
 	};
```
Vfivew, line 60: Now i use only UseEffect to control my url.

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,6 +38,23 @@ const usePagination: UsePagination = ({
 	});
 	const pages = getPagesRange(pagesCut.start, pagesCut.end);
 
+	useEffect(() => {
+		const updatedSearchParameters = new URLSearchParams(
+			searchParameters.toString(),
+		);
+
+		if (pagesCount >= pageFromQuery) {
```
fshabanov, line 46: What if it is a negative page?

fshabanov (changes requested):

YourTeaChan (approved):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
Vfivew: Redirect to 1th page.
redirect

fshabanov (approved):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -14,13 +14,23 @@ type UsePagination = (options: {
 	pagesCount: number;
 };
 
+const QUERY_SECTION_SIZE = 1;
```
what1s1ove, line 17: move to libs in this folder

what1s1ove (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,6 +39,34 @@ const usePagination: UsePagination = ({
...
+			setPage(PaginationValue.DEFAULT_COUNT);
+
+			return;
+		}
+
+		if (pagesCount >= pageFromQuery) {
+			updatedSearchParameters.set(queryName, String(page));
+			setSearchParameters(updatedSearchParameters);
+		} else if (
+			pagesCount !== PaginationValue.PAGE_NOT_EXISTS ||
```
what1s1ove, line 58: move this condition to variable

what1s1ove (changes requested):

apps/frontend/src/libs/hooks/use-pagination/libs/enums/query-section.enum.ts
```diff
@@ -0,0 +1,5 @@
+const QuerySection = {
```
Oleksandra-Masiuk, line 1: Why is it an enum? Will it have more properties, or is it meant to be used as a constant for checking conditions? 
Also, why is it named `QuerySection` and `SIZE`? From the name alone, its purpose isn't clear. Is it meant as a minimum or maximum query length for pagination? Please rename

Oleksandra-Masiuk (changes requested):

Ann-Maksay (approved):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -29,6 +38,35 @@ const usePagination: UsePagination = ({
 	});
 	const pages = getPagesRange(pagesCut.start, pagesCut.end);
 
+	useEffect(() => {
+		const isValidPage = pagesCount >= pageFromQuery;
+		const isInvalidPage =
+			pagesCount !== PaginationValue.PAGE_NOT_EXISTS ||
+			queryParameters.size > QUERY_SECTION_MAX_SIZE;
```
fshabanov, line 45: Why do we need to check the size of query params? Maybe we will have other params?

fshabanov (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -19,8 +19,15 @@ const usePagination: UsePagination = ({
 	pagesCutCount,
 	totalCount,
 }) => {
-	const [page, setPage] = useState<number>(PaginationValue.DEFAULT_PAGE);
-
+	const queryName = "page";
```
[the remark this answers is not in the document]
Vfivew: So, as part of this task, I have to add another hook for params?
Or should I just do it like in the screenshot for now?
queryname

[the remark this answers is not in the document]
what1s1ove: as on screenshot

what1s1ove (approved):

PR #404, opened 2024-03-07
feat(backend/frontend/shared): add real time notifications tm-260

Sytnikovakatya:
Implement real-time notifications


packages/shared/src/modules/notifications/libs/types/notification-with-receiver-id-response-dto.type.ts
```diff
@@ -0,0 +1,6 @@
+import { type NotificationResponseDto } from "./notification-response-dto.type.js";
+
+type NotificationWithReceiverIdResponseDto = Record<"receiverId", number> &
+	Record<"notification", NotificationResponseDto>;
```
fshabanov, line 4:
```suggestion
type NotificationWithReceiverIdResponseDto = {
  receiverId: number;
	notification: NotificationResponseDto>;
}
```
Why not like this?

fshabanov (changes requested):

Oleksandra-Masiuk (changes requested):

GEOFARL (approved):

apps/backend/src/libs/modules/socket/socket-service.module.ts
```diff
@@ -17,6 +17,16 @@ class SocketService {
 		});
 	}
 
+	private notificationsHandler(socket: Socket): void {
```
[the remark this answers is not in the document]
Sytnikovakatya: Yes, sure

Oleksandra-Masiuk (approved):

apps/backend/src/modules/friends/friend.service.ts
```diff
@@ -52,12 +60,19 @@ class FriendService {
...
 			userId: followerUserId,
 		});
 
+		this.socketService.emitMessage({
+			event: SocketEvent.NOTIFICATIONS_NEW_FOLLOWER,
+			payload: notification,
+			receiversIds: [String(notification.userId), String(receiverId)],
+			targetNamespace: SocketNamespace.NOTIFICATIONS,
+		});
+
```
MappariChan, line 75: I think you should do this inside of the notification service and inject this socket service into it

packages/shared/src/libs/modules/socket/libs/enums/socket-event.enum.ts
```diff
@@ -3,6 +3,9 @@ const SocketEvent = {
 	CHAT_JOIN_ROOM: "chat-join-room",
 	CHAT_LEAVE_ROOM: "chat-leave-room",
 	CONNECTION: "connection",
+	NOTIFICATIONS_JOIN_ROOM: "notifications-join-room",
+	NOTIFICATIONS_LEAVE_ROOM: "notifications-leave-room",
+	NOTIFICATIONS_NEW_FOLLOWER: "notifications-new-follower",
```
MappariChan, line 8:
```suggestion
	NEW_NOTIFICATION: "new-notification",
```
I think there will be only one socket message handler for all type of notifications, so I think this enum value would be better

apps/frontend/src/libs/modules/store/middlewares/notifications-socket.middleware.ts
```diff
@@ -0,0 +1,60 @@
...
+	return ({ dispatch }) => {
+		notificationsSocketInstance.on(
+			SocketEvent.NOTIFICATIONS_NEW_FOLLOWER,
+			() => {
+				void dispatch(userNotificationsActions.getUserNotifications());
+				void dispatch(
+					userNotificationsActions.checkHasUserUnreadNotifications(),
+				);
+			},
+		);
```
MappariChan, line 30: I think you should retrieve notification from the payload of the socket message and set it to the state of the notification slice, but not sending new request on the data

apps/backend/src/modules/friends/friend.service.ts
Sytnikovakatya: Ok, I will move it to the notification service

PR #267, opened 2024-03-01
feat(frontend): add course description screen tm-28

Sytnikovakatya:
- Fetch the data for a selected course and sections of the selected course
- Create a screen on frontend and display course data

Screenshot_6
Screenshot_7

PR #407, opened 2024-03-07
feat(frontend/shared): Update the design of the "Chats" page tm-388

GEOFARL:
image
image
image
image
image
image


apps/frontend/package.json
```diff
@@ -21,11 +21,13 @@
 		"@reduxjs/toolkit": "2.0.1",
 		"@trackmates/shared": "*",
 		"clsx": "2.1.0",
+		"date-fns": "3.3.1",
```
Ann-Maksay, line 24: I think all helpers that use `date-fns` should be in `shared`, we already have that package there

GEOFARL: But does it make sense to put functions in the shared folder that will only be used on the frontend?

Ann-Maksay: I guess [this](https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/267#discussion_r1508827432) is also only used on the frontend, so it doesn't really matter

GEOFARL: Thanks, i'll move it to shared

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
```diff
@@ -1,9 +1,16 @@
+import { Fragment } from "react";
+
 import defaultAvatar from "~/assets/img/default-avatar.png";
-import { Image, Loader } from "~/libs/components/components.js";
+import { Button, Image, Loader } from "~/libs/components/components.js";
+import { LAST_ARRAY_ITEM } from "~/libs/constants/constants.js";
+import { checkIsDatePrecedesAnotherByOneDay } from "~/libs/helpers/helpers.js";
+import { useAppDispatch, useCallback } from "~/libs/hooks/hooks.js";
 import { type ChatMessageItemResponseDto } from "~/modules/chat-messages/chat-messages.js";
+import { actions } from "~/modules/chats/chats.js";
```
what1s1ove, line 9: rename

apps/frontend/package.json
```diff
@@ -26,6 +26,7 @@
 		"react-hook-form": "7.49.3",
 		"react-redux": "9.1.0",
 		"react-router-dom": "6.21.3",
+		"react-textarea-autosize": "8.5.3",
```
what1s1ove, line 29:
```suggestion
```
use just a html tag

apps/frontend/src/libs/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const Textarea = <T extends FieldValues>({
+	className,
+	control,
+	errors,
+	hasVisuallyHiddenLabel,
+	label,
+	name,
+	placeholder = "",
+	rows,
```
what1s1ove, line 33: you do not have this component. you need to add a `rows` prop to the `Input` component. And if this props is passed you need to render a `textarea`

packages/shared/src/libs/helpers/get-formatted-chat-date/get-formatted-chat-date.ts
```diff
@@ -0,0 +1,17 @@
+import { format, isThisYear, isToday, isYesterday } from "date-fns";
+
+const getFormattedChatDate = (dateString: string): string => {
```
what1s1ove, line 3:
```suggestion
```
you need to use get formatted date instead. this is a frontend helper only, so move it there

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
```diff
@@ -21,28 +28,77 @@ const Chat: React.FC<Properties> = ({
...
-						isCurrentUserSender={receiver.id !== message.senderUser.id}
-						key={message.id}
-						message={message}
+				{messages.map((message) => {
+					if (checkIsDatePrecedesAnotherByOneDay(message.createdAt, lastDate)) {
+						const previousDate = lastDate;
+						lastDate = message.createdAt;
+
+						return (
+							<Fragment key={message.id}>
```
what1s1ove, line 74: create a new component for message item

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
```diff
@@ -21,28 +28,77 @@ const Chat: React.FC<Properties> = ({
 	onSubmit,
 	receiver,
 }: Properties) => {
+	const dispatch = useAppDispatch();
+
+	const handleClick = useCallback((): void => {
+		dispatch(actions.leaveChat());
+	}, [dispatch]);
+
+	let lastDate = "";
```
what1s1ove, line 37:
```suggestion
```
bad solution. try to find a better solution

what1s1ove (changes requested):

packages/shared/src/libs/helpers/get-formatted-chat-date/get-formatted-chat-date.ts
GEOFARL: The getFormattedDate fn that we already have right now does not produce the result that we need to have in the chat window, needed to create a separate helper. Moved it to the shared folder, to not install 'date-fns' package on the frontend side as was pointed out [here](https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/407#discussion_r1516854005)

apps/frontend/package.json
GEOFARL: image
It just makes it possible for native textarea element to grow according to the amount of content

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
GEOFARL: We already have a component for message `ChatMessage`. In this case `ChatDate` is used to add date delimiter when a message starts from a new date
image

Do not think that it makes sense to create a new component out of it

packages/shared/src/libs/helpers/get-formatted-chat-date/get-formatted-chat-date.ts
Ann-Maksay: What is meant here is that you should leave all date-fsn-related functions in shared in wrappers and combine them in getFormattedChatDate on the frontend.  Take a look at the getFormattedChatDate helper in shared.
Sorry if I didn't explain it clearly enough in my comment

apps/frontend/package.json
what1s1ove: give the static height. it will be enough

what1s1ove (changes requested):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -59,16 +66,36 @@ const Input = <T extends FieldValues>({
 		hasVisuallyHiddenLabel && "visually-hidden",
 	);
 
+	const handleKeyDown = useCallback(
```
what1s1ove, line 69:
```suggestion
```
without any specific behaviour

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -21,7 +22,9 @@ type Properties<T extends FieldValues> = {
 	iconName?: IconName;
 	label: string;
 	name: FieldPath<T>;
+	onSubmit?: (_event: React.BaseSyntheticEvent) => void;
```
what1s1ove, line 25:
```suggestion
```
textereat shouldn't know anything about submit

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -22,12 +22,27 @@
...
 
-.input::placeholder {
+.input::placeholder,
+.textarea::placeholder {
 	font-weight: 400;
 	color: var(--color-menu);
 	opacity: 0.6;
 }
 
+.textarea {
```
what1s1ove, line 32: you can use the same class name as for the input

apps/frontend/src/libs/components/input/input.tsx
what1s1ove: you do not need to handle it at all this time

packages/shared/src/libs/helpers/helpers.ts
```diff
@@ -6,3 +7,4 @@ export { getSanitizedHtml } from "./get-sanitized-html/get-sanitized-html.helper
 export { getSizeInBytes } from "./get-size-in-bytes/get-size-in-bytes.helper.js";
 export { getTimeDistanceFormatDate } from "./get-time-distance-format-date/get-time-distance-format-date.helper.js";
 export { initDebounce } from "./init-debounce/init-debounce.helper.js";
+export { isThisYear, isToday, isYesterday } from "date-fns";
```
what1s1ove, line 10:
```suggestion
```
create yourown wrappers for these functions

packages/shared/src/libs/constants/constants.ts
```diff
@@ -1,2 +1,4 @@
 export { EMPTY_ARRAY_LENGTH } from "./empty-array-length.constant.js";
+export { LAST_ARRAY_ITEM } from "./last-array-item.js";
```
YourTeaChan, line 2:
```suggestion
export { LAST_ARRAY_ITEM } from "./last-array-item.constant.js";
```

packages/shared/src/libs/constants/constants.ts
```diff
@@ -1,2 +1,4 @@
...
 export { MAX_FILE_SIZE_IN_MB } from "./max-file-size.constant.js";
+export { PREVIOUS_INDEX_OFFSET } from "./previous-index-offset.js";
```
YourTeaChan, line 4:
```suggestion
export { PREVIOUS_INDEX_OFFSET } from "./previous-index-offset.constant.js";
```

packages/shared/src/libs/helpers/helpers.ts
```diff
@@ -1,3 +1,4 @@
+export { checkIsDatePrecedesAnotherByOneDay } from "./check-is-date-precedes-another-by-one-day/check-is-date-precedes-another-by-one-day.js";
```
YourTeaChan, line 1:
```suggestion
export { checkIsDatePrecedesAnotherByOneDay } from "./check-is-date-precedes-another-by-one-day/check-is-date-precedes-another-by-one-day.helper.js";
```

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
```diff
@@ -21,28 +29,78 @@ const Chat: React.FC<Properties> = ({
...
 			</div>
 			<ul className={styles["chat-container"]}>
-				{messages.map((message) => (
-					<ChatMessage
-						isCurrentUserSender={receiver.id !== message.senderUser.id}
-						key={message.id}
-						message={message}
+				{messages.flatMap((message, index) => {
+					const components = [];
+					const previousMessage = messages[index - PREVIOUS_INDEX_OFFSET];
```
what1s1ove, line 69: it's really strange logic here. why we can't just map everything?

in render we should not have such logic

move calculation of messages (not components renders) to helper and then use it here

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
```diff
@@ -1,9 +1,17 @@
 import defaultAvatar from "~/assets/img/default-avatar.png";
-import { Image, Loader } from "~/libs/components/components.js";
+import { Button, Image, Loader } from "~/libs/components/components.js";
+import {
+	LAST_ARRAY_ITEM,
+	PREVIOUS_INDEX_OFFSET,
+} from "~/libs/constants/constants.js";
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
```
what1s1ove, line 7: import

what1s1ove (changes requested):

packages/shared/src/libs/helpers/helpers.ts
fshabanov, line 10: Rename to be compatible with B20

apps/frontend/src/pages/chats/libs/components/chat-date/chat-date.tsx
```diff
@@ -0,0 +1,19 @@
+import { getFormattedChatDate } from "../../../libs/helpers/helpers.js";
```
fshabanov, line 1: Fix import path

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
fshabanov, line 7:
```suggestion
import { AppRoute } from "~/libs/enums/enums.js";
```

fshabanov (changes requested):

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -21,7 +21,9 @@ type Properties<T extends FieldValues> = {
 	iconName?: IconName;
 	label: string;
 	name: FieldPath<T>;
+	onKeyDown?: (_event: React.KeyboardEvent) => void;
```
what1s1ove, line 24:
```suggestion
```
you do not to handle any specific keys. remove this handler from common components

apps/frontend/src/libs/enums/event-code.enum.ts
```diff
@@ -0,0 +1,5 @@
+const EventCode = {
```
what1s1ove, line 1:
```suggestion
```
do not handle this this time

apps/frontend/src/pages/chats/libs/helpers/helpers.ts
```diff
@@ -0,0 +1,2 @@
+export { getFormattedChatDate } from "./get-formatted-chat-date.helper.js";
```
what1s1ove, line 1: each helper should be in a separate folder

what1s1ove (changes requested):

what1s1ove (approved):
@GEOFARL resolve comments please

apps/frontend/src/pages/chats/libs/helpers/get-formatted-chat-date/get-formatted-chat-date.helper.ts
```diff
@@ -0,0 +1,26 @@
...
+	} else if (checkIsYesterday(date)) {
+		return DateValue.YESTERDAY;
+	} else if (checkIsThisYear(date)) {
+		return getFormattedDate(dateString, FormatDateType.MMMM_DO).toLowerCase();
+	} else {
+		return getFormattedDate(
+			dateString,
+			FormatDateType.MMMM_DO_YYYY,
+		).toLowerCase();
+	}
```
Oleksandra-Masiuk, line 23:
```suggestion
	if (checkIsToday(date)) {
		return DateValue.TODAY;
	} 
	if (checkIsYesterday(date)) {
		return DateValue.YESTERDAY;
	}
	if (checkIsThisYear(date)) {
		return getFormattedDate(dateString, FormatDateType.MMMM_DO).toLowerCase();
	} 
	return getFormattedDate(
			dateString,
			FormatDateType.MMMM_DO_YYYY,
		).toLowerCase();
```

apps/frontend/src/pages/chats/libs/helpers/prepare-message-items/prepare-message-items.helper.ts
```diff
@@ -0,0 +1,35 @@
...
+const prepareMessageItems = (
+	messages: ChatMessageItemResponseDto[],
+): MessageItem[] => {
+	const items = [];
+
+	for (const [index, message] of messages.entries()) {
+		const previousMessage = messages[index - PREVIOUS_INDEX_OFFSET];
+
+		if (
+			previousMessage &&
```
Oleksandra-Masiuk, line 17: Could you please, move this condition to a separate variable with a descriptive name?

Oleksandra-Masiuk (changes requested):

apps/frontend/src/pages/chats/libs/components/chat/chat.tsx
```diff
@@ -21,28 +27,62 @@ const Chat: React.FC<Properties> = ({
...
+							alt="User avatar"
+							height="48"
+							shape="circle"
+							src={receiver.avatarUrl ?? defaultAvatar}
+							width="48"
+						/>
+					</div>
+					<span
+						className={styles["full-name"]}
+					>{`${receiver.firstName} ${receiver.lastName}`}</span>
```
fshabanov, line 61:
```suggestion
					>{receiver.firstName} {receiver.lastName}</span>
```

fshabanov (changes requested):

what1s1ove (approved):

PR #423, opened 2024-03-08
feat(frontend): add empty page placeholders tm-398

tarmiel:
Screenshot_2
изображение


apps/frontend/src/pages/friends/friends.tsx
```diff
@@ -104,6 +115,7 @@ const Friends: React.FC = () => {
 			case AppRoute.FRIENDS: {
 				return (
 					<FriendsTab
+						emptyPlaceholder={"There aren't any potential friends"}
```
fshabanov, line 118:
```suggestion
						emptyPlaceholder="There aren't any potential friends"
```

fshabanov (changes requested):

apps/frontend/src/pages/feed/feed.tsx
```diff
@@ -21,14 +22,23 @@ const Feed: React.FC = () => {
...
 				<Loader color="orange" size="large" />
 			) : (
-				<FeedActivityList activities={activities} />
+				<>
+					{hasActivities ? (
+						<FeedActivityList activities={activities} />
+					) : (
+						<p className={styles["empty-page-placeholder"]}>
+							There are no activities yet
+						</p>
```
fshabanov, line 39: I see you use the same thing in multiple places, maybe we should create a separate component for that?

fshabanov (changes requested):

apps/frontend/src/pages/feed/feed.tsx
tarmiel: That's actually what I thought at first, too. But technically speaking, this text can change on different pages when adapting or we can add pictures. Right now I've made it almost the same because at this stage it's enough, but I suppose that could change.

Vfivew (approved):

GEOFARL (approved):

apps/frontend/src/pages/feed/feed.tsx
fshabanov: You can send text as a prop. In case we decide to add pictures, we can use props again

what1s1ove (approved):
@fshabanov merge it once it done

fshabanov (approved):

PR #424, opened 2024-03-08
feat(backend/frontend): Add the "Search" field to the header on the "Notifications" page tm-293

GEOFARL:


apps/frontend/src/libs/components/search-bar/search-bar.tsx
```diff
@@ -30,9 +33,26 @@ const SearchBar: React.FC<Properties> = ({
 		user: state.auth.user as UserAuthResponseDto,
 	}));
 	const dispatch = useAppDispatch();
+	const { pathname } = useLocation();
+
+	const SearchHandler = {
```
what1s1ove, line 38: it should be map data structure

btw, can we use switch/case here? if not, use a if + return for not

apps/frontend/src/libs/components/search-bar/libs/maps/search-page-path-to-default-value.map.ts
```diff
@@ -0,0 +1,9 @@
+import { AppRoute } from "~/libs/enums/enums.js";
+import { DEFAULT_SEARCH_MY_COURSES_PAYLOAD } from "~/modules/courses/courses.js";
+
+const SearchPagePathToDefaultValue = {
```
what1s1ove, line 4: > A map is a very useful and frequently used data structure in any programming language. For maps in the JavaScript world, there is a specific naming convention. The name should follow the pattern aToB, where a serves as the key for retrieving values from the map, followed by the preposition To, implying the mapping of two things, and then B, which is the mapped value for a.

https://dev.to/what1s1ove/javascript-naming-conventions-are-important-45n

what1s1ove (changes requested):

apps/backend/src/modules/user-notifications/user-notification.controller.ts
```diff
@@ -149,12 +154,14 @@ class UserNotificationController extends BaseController {
 	 */
 	public async getNotificationsByUserId(
 		options: APIHandlerOptions<{
+			query?: { search?: string };
```
tarmiel, line 157: Why do you need query to be optional?

apps/frontend/src/libs/components/search-bar/search-bar.tsx
```diff
@@ -30,9 +30,10 @@ const SearchBar: React.FC<Properties> = ({
 		user: state.auth.user as UserAuthResponseDto,
 	}));
 	const dispatch = useAppDispatch();
+	const { pathname } = useLocation();
 
 	const { control, errors, handleSubmit } = useAppForm({
-		defaultValues: DEFAULT_SEARCH_MY_COURSES_PAYLOAD,
+		defaultValues: searchPagePathToDefaultValue[pathname as SearchPagePath],
```
what1s1ove, line 36:
```suggestion
		defaultValues: searchPagePathToDefaultValue[pathname as keyof typeof searchPagePathToDefaultValue],
```

what1s1ove (changes requested):

what1s1ove (approved):

PR #427, opened 2024-03-09
feat(backend): Implement cascade user deletion tm-402

GEOFARL:


apps/backend/src/db/migrations/20240309063712_add_notifications_cascade_deletion.ts
```diff
@@ -0,0 +1,36 @@
...
+
+async function up(knex: Knex): Promise<void> {
+	await knex.schema.alterTable(TableName.CHAT_MESSAGES, (table) => {
+		table.dropForeign(ColumnName.USER_ID);
+
+		table
+			.foreign(ColumnName.USER_ID)
+			.references(`${TableName.USERS}.${ColumnName.ID}`)
+			.onDelete(DELETE_STRATEGY);
+	});
```
fshabanov, line 23: I think the existing notifications will now have `userId: null`, you should handle this

fshabanov (changes requested):

what1s1ove (approved):
@fshabanov merge it once it's done

apps/backend/src/db/migrations/20240309063712_add_notifications_cascade_deletion.ts
GEOFARL: I cannot reproduce such behavior, I created notifications before applying migration, then migrated, deleted the user and notifications were gone. Can you explain why do you think the userId will be null?

apps/backend/src/db/migrations/20240309063712_add_notifications_cascade_deletion.ts
```diff
@@ -0,0 +1,36 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	CHAT_MESSAGES: "notifications",
```
fshabanov, line 4:
```suggestion
	NOTIFICATIONS: "notifications",
```

fshabanov (changes requested):

fshabanov (approved):

PR #431, opened 2024-03-09
feat(backend): add users own activity tm-420

Ann-Maksay:
own-activitis


apps/backend/src/modules/activities/activity.repository.ts
```diff
@@ -111,24 +111,7 @@ class ActivityRepository implements Repository<ActivityEntity> {
...
-										id: permission.id,
-										key: permission.key,
-										name: permission.name,
-										updatedAt: permission.updatedAt,
-									});
-								}),
-								updatedAt: group.updatedAt,
-							});
-						}),
+						groups: [],
```
Ann-Maksay, line 114: I was a little stuck trying to figure out how this situation is different from the ones where everything is working correctly. I figured, as we discussed, that we don't need to return this information anyway, but I'm still screwing around trying to figure out the difference.

Specifically, the activity query returns such a user, which caused the map method to fail:
group-1

When in work situations, this one
group-2

what1s1ove: investigate this. we can not delete this code just because

what1s1ove (changes requested):

apps/backend/src/modules/activities/activity.repository.ts
YourTeaChan: If it's still a problem, try to add a relation `${RelationName.GROUPS}.${RelationName.PERMISSIONS}` in `withGraphJoined()` where you need to do `groups.map()`

Vfivew (approved):

YourTeaChan (approved):

ShadowOfTheSpace (approved):

fshabanov (approved):

PR #448, opened 2024-03-09
feat(backend/frontend/shared): filter notifications by type tm-288

ShadowOfTheSpace:


apps/frontend/src/modules/user-notifications/user-notificaitons-api.ts
```diff
@@ -37,19 +40,31 @@ class UserNotificationsApi extends BaseHTTPApi {
...
+
+		options.query = {};
+
+		if (type) {
+			options.query["type"] = type;
+		}
+
+		if (search) {
+			options.query["search"] = search;
+		}
```
tarmiel, line 63: Do you really need this logic? You can set params by default, anyway you are doing it on backend

apps/backend/src/modules/notifications/notification.repository.ts
```diff
@@ -137,19 +140,26 @@ class NotificationRepository implements Repository<NotificationEntity> {
...
-				void builder
-					.whereILike("user:userDetails.firstName", `%${search}%`)
-					.orWhereILike("user:userDetails.lastName", `%${search}%`)
-					.orWhereRaw("concat(??, ' ', ??) ILIKE ?", [
-						"user:userDetails.firstName",
-						"user:userDetails.lastName",
-						`%${search}%`,
-					]);
+			.andWhere((builder) => {
+				if (search) {
```
fshabanov, line 149: No need to have this check, if `search` is empty, `%%` will return everything

apps/backend/src/modules/notifications/notification.repository.ts
```diff
@@ -137,19 +140,26 @@ class NotificationRepository implements Repository<NotificationEntity> {
...
+						.whereILike("user:userDetails.firstName", `%${search}%`)
+						.orWhereILike("user:userDetails.lastName", `%${search}%`)
+						.orWhereRaw("concat(??, ' ', ??) ILIKE ?", [
+							"user:userDetails.firstName",
+							"user:userDetails.lastName",
+							`%${search}%`,
+						]);
+				}
+
+				if (type) {
```
fshabanov, line 160: We will always have type here, because in controller you set a default value

apps/frontend/src/libs/modules/store/middlewares/notifications-socket.middleware.ts
```diff
@@ -20,7 +20,12 @@ const notificationsSocket = ({
 
 	return ({ dispatch }) => {
 		notificationsSocketInstance.on(SocketEvent.NEW_NOTIFICATION, () => {
-			void dispatch(userNotificationsActions.getUserNotifications());
+			void dispatch(
+				userNotificationsActions.getUserNotifications({
+					search: "",
+					type: "all",
```
fshabanov, line 26:
```suggestion
					type: NotificationFilter.ALL,
```

apps/frontend/src/pages/notifications/notifications.tsx
```diff
@@ -1,18 +1,42 @@
...
 
 import { NotificationList } from "./libs/components/notification-list/notification-list.js";
 import styles from "./styles.module.css";
 
 const Notifications: React.FC = () => {
+	const navigate = useNavigate();
+	const [queryParameters] = useSearchParams();
+	const notificationType = queryParameters.get("type") as ValueOf<
+		typeof NotificationFilter
+	> | null;
```
fshabanov, line 34: This is an invalid cast, we can't be sure what we will have in query

fshabanov (changes requested):

apps/backend/src/modules/notifications/notification.repository.ts
ShadowOfTheSpace: Yes, you are right. Search %% returns all results. But I did some investigation about this and wrote comment to you in one of previous PRs. But unfortunately it was merged without any answer. Can you check it now? It is in this [PR]( https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/366#discussion_r1514658550).

apps/backend/src/modules/notifications/notification.repository.ts
ShadowOfTheSpace: Here I need this check because I have 3 types of notification (`new-comment`, `new-follower`, `new-like`). But in case of receiving `all` notifications, I don't need to filter at all. For that case I created in map `filterQueryParameterToNotificationType` field `ALL` with value of `empty string`. If I need to get `all` notifications - filter will be `""` and at repository level I will understand, that I don't need to filter by type at all.
Is it possible to leave this if statement here or I need to change filtering logic?

Vfivew (approved):

YourTeaChan (approved):

apps/backend/src/modules/notifications/notification.repository.ts
fshabanov: Understood, I missed the part with the mapper
Since we will not show multiple notification types together (except the case with ALL notifications), I think there's no need to have an array here

apps/backend/src/modules/notifications/notification.repository.ts
fshabanov: That's a good point, if it really affects the performance that much, we can have a condition there

packages/shared/src/modules/notifications/libs/enums/notification-type.enum.ts
```diff
@@ -1,5 +1,7 @@
 const NotificationType = {
+	NEW_COMMENT: "new-comment",
```
what1s1ove, line 2: we should add new values to db enum as well

apps/frontend/src/pages/notifications/notifications.tsx
```diff
@@ -27,15 +49,57 @@ const Notifications: React.FC = () => {
...
 		<div className={styles["page"]}>
 			<div className={styles["content"]}>
-				<h2 className={styles["title"]}>Notifications</h2>
+				<div className={styles["header"]}>
+					<h2 className={styles["title"]}>Notification</h2>
+					<div className={styles["filters"]}>
+						{possibleTypeValues.map((filter) => {
+							const currentLink =
+								AppRoute.NOTIFICATIONS +
+								(filter === NotificationFilter.ALL ? "" : `?type=${filter}`);
```
what1s1ove, line 83: use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams to work with query

apps/frontend/src/pages/notifications/notifications.tsx
```diff
@@ -1,18 +1,40 @@
...
+import { type ValueOf } from "~/libs/types/types.js";
 import { actions } from "~/modules/user-notifications/user-notifications.js";
 
 import { NotificationList } from "./libs/components/notification-list/notification-list.js";
 import styles from "./styles.module.css";
 
 const Notifications: React.FC = () => {
+	const navigate = useNavigate();
+	const [queryParameters] = useSearchParams();
+	const notificationType = queryParameters.get("type");
```
what1s1ove, line 32: add constant/enum for `type `

apps/frontend/src/modules/user-notifications/user-notificaitons-api.ts
```diff
@@ -37,7 +42,10 @@ class UserNotificationsApi extends BaseHTTPApi {
 		return await response.json<boolean>();
 	}
 
-	public async getUserNotifications(search: string): Promise<{
+	public async getUserNotifications(
```
what1s1ove, line 45: should be an object with params

what1s1ove (changes requested):

Sytnikovakatya (approved):

apps/frontend/src/modules/user-notifications/user-notificaitons-api.ts
Vfivew: So, as part of this task, I have to add another hook for params?
Or should I just do it like in the screenshot for now?
queryname

MappariChan (approved):

apps/backend/src/db/migrations/20240311090319_add_notification_comment_and_like_types.ts
```diff
@@ -0,0 +1,43 @@
...
+	});
+
+	await knex.schema.alterTable(TABLE_NAME, (table) => {
+		table
+			.string(COLUMN_NAME)
+			.checkIn([NotificationType.NEW_FOLLOWER], CHECK_NAME)
+			.notNullable()
+			.alter();
+	});
+}
```
fshabanov, line 41: After down the DB should be as it was before, use table.enum

apps/frontend/src/pages/notifications/libs/constants/query-parameter-name.constant.ts
```diff
@@ -0,0 +1,3 @@
+const QUERY_PARAMETER_NAME = "type";
```
fshabanov, line 1:
```suggestion
const NOTIFICATION_TYPE_QUERY_PARAMETER_NAME = "type";
```
We can have lots of query parameters

fshabanov (changes requested):

apps/backend/src/db/migrations/20240311090319_add_notification_comment_and_like_types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+async function up(knex: Knex): Promise<void> {
+	await knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.dropChecks(CHECK_NAME);
+	});
+
+	await knex.schema.alterTable(TABLE_NAME, (table) => {
+		table
+			.string(COLUMN_NAME)
+			.checkIn(Object.values(NotificationType), CHECK_NAME)
```
what1s1ove, line 23: should be an enum

apps/frontend/src/pages/notifications/libs/constants/query-parameter-name.constant.ts
what1s1ove, line 1: should be an enum

what1s1ove (changes requested):

GEOFARL (approved):

apps/frontend/src/modules/user-notifications/user-notificaitons-api.ts
```diff
@@ -37,7 +42,13 @@ class UserNotificationsApi extends BaseHTTPApi {
 		return await response.json<boolean>();
 	}
 
-	public async getUserNotifications(search: string): Promise<{
+	public async getUserNotifications({
+		search = "",
```
Oleksandra-Masiuk, line 46: Default parameters (`search`) should be last

Oleksandra-Masiuk (changes requested):

apps/frontend/src/modules/user-notifications/user-notificaitons-api.ts
ShadowOfTheSpace: I did this, but we have a linter, which demands sorting of object keys.
image
image

PR #449, opened 2024-03-10
feat(backend/frontend/shared): Send notification through PWA tm-442

GEOFARL:


apps/frontend/public/sw.js
```diff
@@ -0,0 +1,2 @@
+const a = 5;
```
fshabanov, line 1: What's this?

fshabanov (changes requested):

apps/frontend/public/sw.js
GEOFARL: Needed to commit somehow this service worker file, couldn't leave it just empty 🙂

Vfivew: You ignore in knip.config.ts this file right? 
`ignore: ["**/sw.js/**"]`
And after this you still need this file?
What kind of error you got when you try push without this file?

Sytnikovakatya (approved):

apps/frontend/public/sw.js
what1s1ove: we have this file since the PWA vite plugin generates the service worker. remove it

apps/frontend/src/modules/user-notifications/slices/user-notifications.slice.ts
```diff
@@ -61,7 +64,21 @@ const { actions, name, reducer } = createSlice({
 	},
 	initialState,
 	name: "user-notifications",
-	reducers: {},
+	reducers: {
+		sendUnreadNotifications(state): void {
```
what1s1ove, line 68: reduces - is a pure function. remove any side effects

apps/frontend/src/libs/modules/pwa-notification/pwa-notification.module.ts
```diff
@@ -0,0 +1,36 @@
+import { GRANTED_PERMISSION } from "./libs/constants/constants.js";
+
+class PWANotification {
```
what1s1ove, line 3:
```suggestion
class PushNotification {
```
change it everywhere

what1s1ove (changes requested):

apps/frontend/public/sw.js
```diff
@@ -0,0 +1,13 @@
...
+
+self.addEventListener("push", async (e) => {
+	const dataObject = await e.data.json();
+	const notificationTitle = dataObject.message;
+	const notificationOptions = dataObject.options;
+
+	e.waitUntil(
+		self.registration.showNotification(notificationTitle, notificationOptions),
+	);
+});
```
GEOFARL, line 13: Have not found better solution with our Vite PWA plugin then creating a separate service worker and injecting it in precache manifest. That's what official Vite PWA docs suggest [here](https://vite-pwa-org.netlify.app/guide/inject-manifest)

apps/frontend/eslint.config.js
```diff
@@ -10,7 +10,7 @@ let FlatConfig;
 
 /** @type {FlatConfig} */
 const ignoresConfig = {
-	ignores: ["build", "dev-dist"],
+	ignores: ["build", "dev-dist", "public/sw.js"],
```
GEOFARL, line 13: I think that it is needed to be ignored by analogy with other service workers

apps/frontend/package.json
```diff
@@ -41,6 +41,7 @@
 		"react-intersection-observer": "9.8.1",
 		"vite": "5.0.12",
 		"vite-plugin-pwa": "0.18.1",
-		"vite-plugin-svgr": "4.2.0"
+		"vite-plugin-svgr": "4.2.0",
+		"workbox-precaching": "7.0.0"
```
GEOFARL, line 45: This dependency is needed to be able to inject service worker

apps/frontend/src/libs/modules/config/base-config.module.ts
```diff
@@ -23,6 +23,11 @@ class BaseConfig implements Config {
 					typeof AppEnvironment
 				>,
 			},
+			PUSH_NOTIFICATIONS: {
+				PUBLIC_KEY: import.meta.env[
+					"VITE_PUSH_NOTIFICATIONS_PUBLIC_KEY"
+				] as string,
+			},
```
GEOFARL, line 30: Need public key here

> to restrict the validity of a subscription to a specific application server (so, by using VAPID, only your server will be able to send notifications to a subscriber).
[good explanation](https://stackoverflow.com/questions/40392257/what-is-vapid-and-why-is-it-useful)

apps/frontend/src/libs/modules/store/libs/helpers/send-notifications.helper.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const sendNotifications = (notifications: NotificationResponseDto[]): void => {
+	for (const notification of notifications) {
+		if (notification.status === NotificationStatus.UNREAD) {
+			void pushNotification.sendNotification({
+				message: notificationTypeToTitle[notification.type],
+				options: {
+					body: notification.message,
+					icon: notification.userAvatarUrl ?? "",
+				},
```
GEOFARL, line 17: Didn't know what kinds of notifications we will send, included this for now, probably, needs to be changed

apps/frontend/src/libs/modules/store/libs/maps/notification-type-to-title.map.ts
```diff
@@ -0,0 +1,7 @@
+import { NotificationType } from "~/modules/user-notifications/user-notifications.js";
+
+const notificationTypeToTitle = {
+	[NotificationType.NEW_FOLLOWER]: "New follower",
+};
```
GEOFARL, line 5: Expect this map to be extended in the future

apps/frontend/src/libs/modules/store/middlewares/notifications-socket.middleware.ts
```diff
@@ -18,10 +20,20 @@ const notificationsSocket = ({
...
+			);
+
+			const { userNotifications } = getState();
+
+			if (!userNotifications.hasUnreadNotifications) {
+				return;
+			}
+
+			sendNotifications(userNotifications.notifications);
 		});
```
GEOFARL, line 37: Probably, we don't need to send out all kinds of notifications to all of the subscribers of our app, left it so for now

apps/frontend/vite.config.ts
```diff
@@ -66,6 +66,7 @@ const config = ({ mode }: ConfigEnv): ReturnType<typeof defineConfig> => {
 		},
 		outDir: "build",
 		registerType: "autoUpdate",
+		strategies: "injectManifest",
```
GEOFARL, line 69: To extend generated service worker

knip.config.ts
```diff
@@ -1,6 +1,8 @@
 import { type KnipConfig } from "knip";
 
 const config: KnipConfig = {
+	ignore: ["**/sw.js/**"],
+	ignoreDependencies: ["workbox-precaching"],
```
GEOFARL, line 5: knip was complaining about them, was not sure what to do about that

apps/backend/.env.example
```diff
@@ -46,3 +46,10 @@ OPENAI_MODEL=gpt-3.5-turbo-0125
 # EDX
 #
 EDX_URL=https://courses.edx.org/
+
+#
+# PUSH_NOTIFICATIONS
+#
+PUSH_NOTIFICATIONS_PRIVATE_KEY=YqPgqaxzdgDQZPkr3nIzRrt8r98R3YtaSLDObHD4xXo
```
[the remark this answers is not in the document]
GEOFARL: [link](https://pushpad.xyz/blog/web-push-what-is-vapid)
> Basically, when you subscribe a user to the notifications, you can also pass a public key (cryptographic key) to the browser. Then all the notifications will need to be signed with the private key.
> These are the exact steps:
> 1. the website backend generates and stores in its own database a public key and private key (which will be called VAPID keys)
> 2. the website frontend calls PushManager.subscribe(options) ([documentation](https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe)) to subscribe the user to notifications and the options include an applicationServerKey, which is the public key downloaded from server
> 3. if the user accepts to receive the notifications (permission prompt), the browser needs to contact its own push service to generate a new push subscription
> 4. the new push subscription returned by the browser to the website is associated to that VAPID public key
> 5. the push subscription can be used only if you have the VAPID private key: you will need to sign all the future notifications sent from your backend (application server) with the VAPID private key.

PR #464, opened 2024-03-11
fix(frontend): feat add pagination with query tm-307

Vfivew:
added check for bigint and fractions.


apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -24,6 +24,7 @@ const usePagination: UsePagination = ({
 	const [searchParameters, setSearchParameters] = useSearchParams();
 	const pageFromQuery = Number(searchParameters.get(queryName));
 	const validPage =
+		(typeof pageFromQuery === "bigint" || Number.isInteger(pageFromQuery)) &&
 		pageFromQuery > PaginationValue.PAGE_NOT_EXISTS
 			? pageFromQuery
 			: PaginationValue.DEFAULT_PAGE;
```
GEOFARL, line 30: probably, it'll be better to split this, seems huge))

Vfivew: You think it better, to create 3 const which i will check after? Example: is BigInt/isInteger/isValid Page And then check 3 cond for Valid page? 
Or what kind of split do you mean?

GEOFARL:
```
const checkIsPageValid = (page: number) => {
    return typeof pageFromQuery !== "bigint" &&
                Number.isInteger(pageFromQuery) && 
                pageFromQuery > PaginationValue.PAGE_NOT_EXISTS
}

const validPage = checkIsPageValid(pageFromQuery)
			? pageFromQuery
			: PaginationValue.DEFAULT_PAGE;
```

maybe sth like this will work

Vfivew: In fact, I'm not sure which is better.
Maybe someone else will express their opinion.

GEOFARL: in your case `validPage` can be a boolean, if I'm not mistaken
not sure whether it is a problem

Oleksandra-Masiuk: No, `validPage` won't be a boolean.
I would have split the condition into two variables:
for ` (typeof pageFromQuery === "bigint" || Number.isInteger(pageFromQuery))`
and `pageFromQuery > PaginationValue.PAGE_NOT_EXISTS`

Oleksandra-Masiuk (approved):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -23,10 +23,11 @@ const usePagination: UsePagination = ({
 }) => {
 	const [searchParameters, setSearchParameters] = useSearchParams();
 	const pageFromQuery = Number(searchParameters.get(queryName));
-	const validPage =
-		pageFromQuery > PaginationValue.PAGE_NOT_EXISTS
-			? pageFromQuery
-			: PaginationValue.DEFAULT_PAGE;
+	const isPageInteger =
+		typeof pageFromQuery !== "bigint" && Number.isInteger(pageFromQuery);
```
what1s1ove, line 27: i am not sure that we can have bigint here

MappariChan (approved):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
Vfivew: I entered huge values and didn't get the behavior I wanted. But now I realized that the error was not in bigint.
Everything is okay now.
bigint

KrasKa2 (approved):

what1s1ove (approved):

PR #461, opened 2024-03-11
feat(backend/shared): display the activities after completing the course on the activities page tm-442 

Vfivew:

PR #466, opened 2024-03-11
feat(backend): add like notification tm-445

Ann-Maksay:
Updated:

- Like notification:
like-notification-new

- Own like:
like-notification-own-like

- Delete like:

like-notification-remove-like

UPD: with counter
like-notification-counter


Vfivew (approved):

apps/backend/src/modules/activities/activity.service.ts
```diff
@@ -41,17 +50,29 @@ class ActivityService implements Service {
...
 
 		const likeObject = like?.toObject();
+		const targetActivity = await this.activityRepository.find(activityId);
 
-		likeObject?.id
-			? await this.activityLikeRepository.delete(likeObject.id)
-			: await this.activityLikeRepository.create(
-					ActivityLikeEntity.initializeNew({ activityId, userId }),
-				);
+		if (likeObject?.id) {
```
Oleksandra-Masiuk, line 55: what is the point of replacing the ternary operator with `else if`?

GEOFARL: I saw Vlad requested that we do not call methods based on the ternary condition [here](https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/461#discussion_r1519929100)

GEOFARL (approved):

apps/backend/src/modules/activities/activity.service.ts
Ann-Maksay: Also, the logic here has become a little more complicated than just choosing one of two options

what1s1ove (changes requested):
merge main into your branch please

MappariChan (approved):

Ann-Maksay:
I've done the merge, but I'm still working on adding [that](https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/466#discussion_r1521050469).
@what1s1ove

apps/backend/src/db/migrations/20240312100649_add_notification_reference_to_likes.ts
```diff
@@ -0,0 +1,31 @@
+import { type Knex } from "knex";
+
+const TableName = {
+	ACTIVITY_LIKES: "activity_likes",
+	NOTIFICATIONS: "notifications",
+} as const;
+
+const ColumnName = {
+	ID: "id",
+	NOTIFICATIN_ID: "notification_id",
```
GEOFARL, line 10:
```suggestion
	NOTIFICATION_ID: "notification_id",
```

Ann-Maksay:
Our notifications aren't particularly adaptable to deletion, so the best option I found was to create a notification for any likes, but only display to users the ones not personally caused by them.

apps/backend/src/modules/notifications/notification.repository.ts
```diff
@@ -37,6 +38,7 @@ class NotificationRepository implements Repository<NotificationEntity> {
 		const unreadNotifications = await this.notificationModel
 			.query()
 			.where({ receiverUserId: userId })
+			.andWhereNot(`${DatabaseTableName.NOTIFICATIONS}.userId`, userId)
```
GEOFARL, line 41: do we have notifications from ourselves?

apps/backend/src/db/migrations/20240312100649_add_notification_reference_to_likes.ts
```diff
@@ -0,0 +1,31 @@
...
+	ID: "id",
+	NOTIFICATION_ID: "notification_id",
+} as const;
+
+const DELETE_STRATEGY = "CASCADE";
+
+async function up(knex: Knex): Promise<void> {
+	await knex.schema.alterTable(TableName.ACTIVITY_LIKES, (table) => {
+		table
+			.integer(ColumnName.NOTIFICATION_ID)
```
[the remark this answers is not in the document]
Ann-Maksay: We have no other way to get it, and it's needed for deletion along with the deletion of the likes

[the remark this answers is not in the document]
KrasKa2: If only for deleting notification when delete like, I think it is better to add `paylod` or `action_id` field to `notifications` table

KrasKa2 (changes requested):

apps/backend/src/db/migrations/20240312100649_add_notification_reference_to_likes.ts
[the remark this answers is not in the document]
KrasKa2: @fshabanov , what do you think?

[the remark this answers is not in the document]
Ann-Maksay: I understand what you're saying and that this is how it's implemented in activity. I considered this option, but it requires too many modifications to the notifications module, which is something I wanted to avoid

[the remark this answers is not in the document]
KrasKa2: In the other hand, we have already had hight coupling because of using notificationService (and activityService) inside other services instead of using events. So, maybe in this case it is ok

KrasKa2 (approved):

what1s1ove (changes requested):
add DB changes to readme.md

apps/backend/src/db/migrations/20240312100649_add_notification_reference_to_likes.ts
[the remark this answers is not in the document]
ShadowOfTheSpace: I agreed with @KrasKa2, I think it is better to add some `actionId` to payload like it was in SectionStatuses. It can be useful in future tasks without any additional tables. It requires one additional field (maybe it can be nullable) with id of entity that created notification.

Sytnikovakatya (approved):

apps/backend/src/db/migrations/20240312100649_add_notification_reference_to_likes.ts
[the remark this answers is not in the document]
what1s1ove: agreed

what1s1ove (changes requested):

apps/backend/src/modules/notifications/notification.service.ts
```diff
@@ -82,6 +84,34 @@ class NotificationService implements Service {
...
+			actionId,
+			type,
+		);
+
+		if (!notification) {
+			throw new NotificationError({
+				message: ExceptionMessage.NOTIFICATION_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
```
tarmiel, line 101: You use notificationService in Activity. Do you suppose that user who tries to delete his own like or comment should get an error message about someone else notification if we cannot find this notification in db?

apps/backend/src/modules/notifications/notification.service.ts
```diff
@@ -82,6 +84,34 @@ class NotificationService implements Service {
...
+
+		const notificationObject = notification.toObject();
+
+		this.socketService.emitMessage({
+			event: SocketEvent.NEW_NOTIFICATION,
```
tarmiel, line 106: Do you want to send new_notification event before deletion?

packages/shared/src/modules/notifications/libs/types/create-notification-request-dto.type.ts
```diff
@@ -2,6 +2,7 @@ import { type ValueOf } from "../../../../libs/types/types.js";
 import { type NotificationType } from "../enums/enums.js";
 
 type CreateNotificationRequestDto = {
+	actionId: null | number;
```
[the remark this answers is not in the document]
Ann-Maksay: Because we don't have deletion on unfollow and there is no action id there. So this field is nullable

apps/backend/src/modules/notifications/notification.service.ts
Ann-Maksay: I did it without the error from the beginning, but I decided to keep it. I just don't cause deletion of the notification if the user removes his own likes. As a consequence, there can only be an error if the user deletes comments that were created before the deletion was possible. But... I think that's where the error message should be.

apps/backend/src/modules/notifications/notification.service.ts
Ann-Maksay: Yes, because this event just updates the notification list. But I agree that maybe it should be renamed.

what1s1ove (approved):

PR #467, opened 2024-03-11
fix(backend/frontend/shared): Update the design of the "Chats" page tm-388

GEOFARL:
Date format should be "Today", "March 7th", "March 4th"
image

Fix button position
image

 Message background shouldn't depend of the user name and date
image

Messages with line breaks are sent on one row(fixed)
image

Validation message shouldn't display, user can't send empty messages


apps/frontend/src/pages/chats/libs/components/chat-form/styles.module.css
```diff
@@ -16,7 +16,7 @@
 	bottom: 0;
 	justify-self: end;
 	width: 40px;
-	height: 44px;
+	height: 45.23px;
```
Vfivew, line 19: 45.23 ?
I think it is bad any way.

GEOFARL: Icon has absolute positioning and without exact height it looks like this
image

Vfivew: Try 46px height
And delete `justify-self: end;` we dont need it at all (optional)
Gif added. 
But check it in all browser.
this

GEOFARL: thks for justify-end, probably artifact of the past))

Vfivew: .icon {
	position: absolute;
	right: 1;
	bottom: 1;
	justify-self: end;
	width: 40px;
	height: 44px;
	min-height: 30px;
	padding: 0;
	background-color: transparent;
	border: none;
	border-radius: 0 7px 7px 0;
	fill: currentcolor;
}

GEOFARL: image

apps/frontend/src/pages/chats/chats.tsx
```diff
@@ -47,6 +48,10 @@ const Chats: React.FC = () => {
 
 	const onSubmit = useCallback(
 		(payload: typeof DEFAULT_MESSAGE_PAYLOAD): void => {
+			if (payload.message.trim().length === EMPTY_ARRAY_LENGTH) {
```
Oleksandra-Masiuk, line 51: It is a bit unnecessary to compare to `EMPTY_ARRAY_LENGTH`
```suggestion
			if (!payload.message.trim().length) {
```

apps/frontend/src/pages/chats/libs/components/chat-form/styles.module.css
Vfivew: I don't know how this happens for you. All media requests are processed correctly for me.
Any way we need to find solution and delete 45.23px
example

GEOFARL: okay, thanks for help, will try to do it the other way

apps/frontend/src/pages/chats/chats.tsx
GEOFARL: image

Vfivew: You miss "!". Try it.
hereTry

GEOFARL: image

Vfivew (approved):

Oleksandra-Masiuk (approved):

apps/frontend/src/pages/chats/chats.tsx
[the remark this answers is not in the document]
what1s1ove: you need to handle it with validation

what1s1ove (changes requested):

apps/frontend/src/pages/chats/chats.tsx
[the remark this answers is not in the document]
GEOFARL: I was asked to not show an arrow message when the input is empty. That's why I've removed validation rule and added this check.

String is technically an array of chars)) I'll create a separate constant for that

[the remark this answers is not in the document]
GEOFARL: > you need to handle it with validation

image

It was a part of my task to remove that, should I return it back?

packages/shared/src/libs/enums/empty-length.enum.ts
```diff
@@ -0,0 +1,6 @@
+const EmptyLength = {
```
fshabanov, line 1: What about having just a constant EMPTY_LENGTH?

fshabanov (changes requested):

apps/frontend/src/pages/chats/chats.tsx
```diff
@@ -47,6 +48,10 @@ const Chats: React.FC = () => {
 
 	const onSubmit = useCallback(
 		(payload: typeof DEFAULT_MESSAGE_PAYLOAD): void => {
+			if (payload.message.trim().length === EMPTY_LENGTH) {
```
what1s1ove, line 51:
```suggestion
```
is not resolved

what1s1ove (changes requested):

apps/frontend/src/pages/chats/chats.tsx
Vfivew: Then there will be an error in the chat. 
The task is to avoid such a message, as it is not prettily.
Max has already found another solution, but do you agree that the error should not be in the chat, or should it be returned?
here123

what1s1ove (approved):

apps/frontend/src/pages/chats/chats.tsx
what1s1ove: i guess we can use more that 1 validation in this case

PR #474, opened 2024-03-12
feat(frontend/backend/shared): uam delete users tm-232

KrasKa2:
task-232


apps/backend/src/modules/users/user.controller.ts
```diff
@@ -24,6 +29,25 @@ class UserController extends BaseController {
 
 		this.userService = userService;
 
+		this.addRoute({
+			handler: (options) =>
```
Vfivew, line 33: B27

apps/frontend/src/pages/management/libs/enums/management-dialogue-messages.ts
```diff
@@ -0,0 +1,5 @@
+const ManagementDialogueMessages = {
+	DO_YOU_WANT_DELETE_USER: "Are you sure, that you want to delete user",
```
Vfivew, line 2:
```suggestion
	DO_YOU_WANT_DELETE_USER: "Are you sure, that you want to delete user?",
```

Vfivew (changes requested):

apps/frontend/src/pages/management/libs/enums/management-dialogue-messages.ts
KrasKa2: @Vfivew , this is using in this way:

`const modalContent = `${ManagementDialogueMessages.DO_YOU_WANT_DELETE_USER} ${user.firstName} ${user.lastName}?`;
`

apps/backend/src/libs/hooks/check-user-permissions/check-user-permissions.hook.ts
```diff
@@ -11,7 +11,7 @@ import {
 	PermissionMode,
 } from "~/libs/enums/enums.js";
 import { type ValueOf } from "~/libs/types/types.js";
-import { PermissionError } from "~/modules/permissions/permissions.js";
+import { PermissionError } from "~/modules/permissions/libs/exceptions/exceptions.js";
```
fshabanov, line 14:
```suggestion
import { PermissionError } from "~/modules/permissions/permissions.js";
```

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -63,6 +88,52 @@ class UserController extends BaseController {
...
+			id: string;
+		};
+		user: UserAuthResponseDto;
+	}>): Promise<APIHandlerResponse> {
+		const userId = Number(id);
+
+		if (user.id === userId) {
+			throw new UserError({
+				message: UserErrorMessage.FORBIDDEN_DELETING_YOURSELF,
+				status: HTTPCode.FORBIDDEN,
```
fshabanov, line 125: Maybe BAD_REQUEST? Forbidden is more about permissions, but deleting yourself I think is not because of permissions

apps/frontend/src/modules/management/slices/management.slice.ts
```diff
@@ -10,6 +10,7 @@ import {
 	type PermissionResponseDto,
 	actions as permissionsActions,
 } from "~/modules/permissions/permissions.js";
+import { remove } from "~/modules/users/slices/actions.js";
```
fshabanov, line 13:
```suggestion
import { actions as userActions } from "~/modules/users/slices/users.js";
```

apps/frontend/src/pages/management/libs/components/delete-button/delete-button.tsx
```diff
@@ -0,0 +1,79 @@
...
+		setIsModalOpen(false);
+	}, [onClick, setIsModalOpen, user.id]);
+
+	const handleCancel = useCallback(() => {
+		setIsModalOpen(false);
+	}, [setIsModalOpen]);
+
+	const handleModalClose = useCallback(() => {
+		setIsModalOpen(false);
+	}, [setIsModalOpen]);
```
fshabanov, line 43: Those are doing the same thing, you can have only one of them

apps/frontend/src/pages/management/libs/components/users-tab/users-tab.tsx
```diff
@@ -1,27 +1,39 @@
 import { Button } from "~/libs/components/components.js";
+import { DataStatus } from "~/libs/enums/data-status.enum.js";
```
fshabanov, line 2:
```suggestion
import { DataStatus } from "~/libs/enums/enums.js";
```

apps/frontend/src/pages/management/libs/components/users-tab/users-tab.tsx
```diff
@@ -1,27 +1,39 @@
...
+			userToDataStatus: state.management.userToDataStatus,
 			users: state.management.users,
 		};
 	});
 
 	const dispatch = useAppDispatch();
 
+	const handleDeleteUser = useCallback(
+		(userId: number) => {
+			void dispatch(remove(userId));
```
fshabanov, line 32:
```suggestion
			void dispatch(usersActions.remove(userId));
```

MappariChan (approved):

apps/frontend/src/pages/management/libs/components/users-tab/users-tab.tsx
```diff
@@ -39,11 +51,11 @@ const UsersTab: React.FC = () => {
 	const tableData = users.map((user) => {
 		return {
 			delete: (
-				<Button
-					className={styles["icon-button"]}
-					hasVisuallyHiddenLabel
-					iconName="delete"
+				<DeleteButton
+					isLoading={userToDataStatus[user.id] == DataStatus.PENDING}
```
fshabanov, line 55:
```suggestion
					isLoading={userToDataStatus[user.id] === DataStatus.PENDING}
```

fshabanov (changes requested):

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -63,6 +88,52 @@ class UserController extends BaseController {
...
+		user: UserAuthResponseDto;
+	}>): Promise<APIHandlerResponse> {
+		const userId = Number(id);
+
+		if (user.id === userId) {
+			throw new UserError({
+				message: UserErrorMessage.FORBIDDEN_DELETING_YOURSELF,
+				status: HTTPCode.BAD_REQUEST,
+			});
+		}
```
fshabanov, line 127: Should be a part of service

apps/frontend/src/pages/management/libs/components/delete-button/delete-button.tsx
```diff
@@ -0,0 +1,74 @@
...
+	const handleDelete = useCallback(() => {
+		onClick(user.id);
+		setIsModalOpen(false);
+	}, [onClick, setIsModalOpen, user.id]);
+
+	const openModal = useCallback(() => {
+		setIsModalOpen(true);
+	}, [setIsModalOpen]);
+
+	const closeModal = useCallback(() => {
```
fshabanov, line 36:
```suggestion
	const handleOpenModal = useCallback(() => {
		setIsModalOpen(true);
	}, [setIsModalOpen]);

	const handleCloseModal = useCallback(() => {
```

fshabanov (changes requested):

GEOFARL (approved):

apps/frontend/src/pages/management/libs/components/confirmation-modal/confirmation-modal.tsx
```diff
@@ -0,0 +1,45 @@
...
+	content,
+	isOpen,
+	onCancel,
+	onClose,
+	onConfirm,
+}: Properties) => {
+	return (
+		<Modal isOpen={isOpen} onClose={onClose} size="small">
+			<div className={styles["confirm-modal"]}>
+				<div className={styles["confirm-modal-content"]}>{content}</div>
```
fshabanov, line 23: Add text inside of `p` tag (or other text tag)

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -63,6 +87,45 @@ class UserController extends BaseController {
...
+	 *      responses:
+	 *        200:
+	 *          description: Successful operation
+	 *          content:
+	 *            application/json:
+	 *              schema:
+	 *                type: object
+	 *                properties:
+	 *                  success:
+	 *                    type: boolean
```
ShadowOfTheSpace, line 108: Do you return `{ success: boolean }` from service? I think you return only boolean value from service and swagger docs should be changed

ShadowOfTheSpace (changes requested):

apps/frontend/src/pages/management/libs/components/delete-button/delete-button.tsx
ShadowOfTheSpace, line 1: Can we use our Button component here?

apps/backend/src/modules/users/user.controller.ts
KrasKa2: @ShadowOfTheSpace No, I return `{ success: boolean }`

apps/frontend/src/pages/management/libs/components/delete-button/delete-button.tsx
KrasKa2: @ShadowOfTheSpace I am using our Button component inside `DeleteButton`

apps/frontend/src/pages/management/libs/enums/management-dialogue-messages.ts
```diff
@@ -0,0 +1,5 @@
+const ManagementDialogueMessages = {
```
ShadowOfTheSpace, line 1:
```suggestion
const ManagementDialogueMessage = {
```

ShadowOfTheSpace (changes requested):

apps/backend/src/modules/users/user.controller.ts
ShadowOfTheSpace: I got it, you are right!

apps/frontend/src/pages/management/libs/components/delete-button/delete-button.tsx
```diff
@@ -0,0 +1,74 @@
...
+					hasVisuallyHiddenLabel
+					iconName="delete"
+					isDisabled={isDisabled}
+					label={label}
+					onClick={handleOpenModal}
+				/>
+			)}
+			{isModalOpen && (
+				<ConfirmationModal
+					// isOpen={isModalOpen}
```
fshabanov, line 63: Remove or uncomment

apps/frontend/src/pages/management/libs/components/users-tab/users-tab.tsx
```diff
@@ -86,12 +96,19 @@ const UsersTab: React.FC = () => {
...
 						onClick={handleOpenEditModal(user)}
 					/>
-					<Button
+					<DeleteButton
+						isLoading={userToDataStatus[user.id] === DataStatus.PENDING}
+						label="delete"
+						onClick={handleDeleteUser}
+						user={user}
+					/>
+					{/* <Button
```
fshabanov, line 105: Remove or uncomment

fshabanov (changes requested):

apps/backend/src/modules/users/user.service.ts
```diff
@@ -57,6 +58,20 @@ class UserService implements Service {
 		return await this.userRepository.delete(userId);
 	}
 
+	public async deleteUserByAdmin(
+		admin: UserAuthResponseDto,
+		userId: number,
```
YourTeaChan, line 63: If you need only `admin id`, you can pass only `id` prop instead of full object

KrasKa2: That what I did at first, but then decided to make this two params different types, in order to avoid switch places, because the consequences can be sad. Maybe it is better to use object as param:
```
{
   userId: number,
   adminId: number,  
}
```
But I decided, that there are not enough parameters here

apps/frontend/src/pages/management/libs/components/users-tab/users-tab.tsx
```diff
@@ -66,6 +70,22 @@ const UsersTab: React.FC = () => {
 		[currentUser, handleChangeUserGroups],
 	);
 
+	const [isDeleteModalOpen, setDeleteModalOpen] = useState<boolean>(false);
```
GEOFARL, line 73:
```suggestion
	const [isDeleteModalOpen, setIsDeleteModalOpen] = useState<boolean>(false);
```

apps/frontend/src/pages/management/libs/enums/management-dialogue-message.enum.ts
```diff
@@ -1,5 +1,6 @@
 const ManagementDialogueMessage = {
 	DELETE_GROUP: "Are you sure, that you want to delete group",
+	DELETE_USER: "Are you sure, that you want to delete user",
```
GEOFARL, line 3: not sure, but maybe let's add question marks?

Oleksandra-Masiuk: @GEOFARL [here](https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/pull/474#discussion_r1521113681) is the answer

Sytnikovakatya (approved):

Vfivew (approved):

apps/backend/src/modules/users/user.service.ts
YourTeaChan: I think, passing an object with `userId` and `adminId` is better idea, than passing all `admin` object, when you need only `id` from it

apps/frontend/src/modules/users/slices/actions.ts
```diff
@@ -9,6 +9,18 @@ import {
 
 import { name as sliceName } from "./users.slice.js";
 
+const remove = createAsyncThunk<number, number, AsyncThunkConfig>(
+	`${sliceName}/remove`,
+	async (id, { extra }) => {
+		const { notification, userApi } = extra;
+
+		await userApi.remove(id);
```
ShadowOfTheSpace, line 17: Why you don't handle situations when backend returns `success: false`? 

It can be strange when a user in DB is not deleted, but on the frontend, we will have a notification about the successful operation. 
Also, it will be wrong, if we remove this user from the redux store without depending on backend response.

I think it should be changed

ShadowOfTheSpace (changes requested):

apps/frontend/src/modules/users/slices/actions.ts
```diff
@@ -1,14 +1,37 @@
...
+		const { notification, userApi } = extra;
+
+		const { success } = await userApi.remove(id);
+
+		if (success) {
+			notification.success(NotificationMessage.USER_DELETED);
+		} else {
+			throw new UserError({
+				message: UserErrorMessage.USER_DELETION_FAILED,
+				status: HTTPCode.BAD_REQUEST,
```
ShadowOfTheSpace, line 27: Maybe would be better to use `notification.error` here and not to throw server side error on frontend?

ShadowOfTheSpace (changes requested):

apps/frontend/src/modules/users/slices/actions.ts
KrasKa2: @ShadowOfTheSpace then I have to return `null` and handle this type of response in the slice (this complicates logic). If we throw error,  error toast is showing. If it was server-side error, it would be in `backend` package, not in  `shared`

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -63,6 +87,45 @@ class UserController extends BaseController {
...
+		user,
+	}: APIHandlerOptions<{
+		params: {
+			id: string;
+		};
+		user: UserAuthResponseDto;
+	}>): Promise<APIHandlerResponse> {
+		const userId = Number(id);
+
+		const success = await this.userService.deleteUserByAdmin(user, userId);
```
what1s1ove, line 121:
```suggestion
		const success = await this.userService.delete(user, userId);
```
it should be just a delete. you need to handle permissions by the hook

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -63,6 +87,45 @@ class UserController extends BaseController {
...
+
+		return {
+			payload: { success },
```
what1s1ove, line 124:
```suggestion
			payload: true or false,
```

apps/frontend/src/pages/management/libs/components/users-tab/users-tab.tsx
```diff
@@ -104,8 +126,11 @@ const UsersTab: React.FC = () => {
 						className={styles["icon-button"]}
 						hasVisuallyHiddenLabel
 						iconName="delete"
-						isDisabled={!hasPermissionToDelete || isAdmin}
+						isDisabled={!hasPermissionToDelete || isSameUser}
+						isLoading={isDeleting}
 						label={UsersTableHeader.BUTTONS}
+						loaderColor="orange"
+						onClick={handleOpenDeleteModal(user)}
```
what1s1ove, line 133:
```suggestion
						onClick={handleOpenDeleteModal}
```
we do not want to have currying functions

what1s1ove (changes requested):

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -24,6 +33,36 @@ class UserController extends BaseController {
...
+				),
+				checkParametersByUser<{ id: string }>(
+					(parameters: { id: string }, user: UserAuthResponseDto) => {
+						return Number(parameters.id) !== user.id;
+					},
+					new UserError({
+						message: UserErrorMessage.FORBIDDEN_DELETING_YOURSELF,
+						status: HTTPCode.BAD_REQUEST,
+					}),
+				),
```
what1s1ove, line 59: all this logic should be inside the hook. in parameters, you should pass only name of the query param

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -24,6 +33,36 @@ class UserController extends BaseController {
...
+			handler: (options) => {
+				return this.delete(
+					options as APIHandlerOptions<{
+						params: { id: string };
+					}>,
+				);
+			},
+			method: "DELETE",
+			path: UsersApiPath.$ID,
+			preHandler: combineHooks([
```
what1s1ove, line 46:
```suggestion
			preHandler: [
```
you do not need to add an addition hook for it. you need to change the logic of the `preHandler` key to allow it accepts an array

apps/frontend/src/pages/management/libs/components/item-button/item-button.tsx
what1s1ove, line 1: it a bad to have this component. it would be better to create an `ActionsCell` where you will have all your handlers

what1s1ove (changes requested):

what1s1ove (approved):

PR #483, opened 2024-03-12
fix(frontend): The "Follow" / "Following" buttons on user components are not updated on the new tab after following / unfollowing the users tm-444

GEOFARL:

### DIFF
--- a/apps/frontend/src/modules/friends/slices/friends.slice.ts
+++ b/apps/frontend/src/modules/friends/slices/friends.slice.ts
@@ -21,6 +21,7 @@ type State = {
 	followings: UserAuthResponseDto[];
 	followingsTotalCount: number;
 	isFollowing: boolean;
+	locallyUnfollowedFriendIds: number[];
 	potentialFriends: UserAuthResponseDto[];
 	potentialFriendsTotalCount: number;
 };
@@ -33,6 +34,7 @@ const initialState: State = {
 	followings: [],
 	followingsTotalCount: 0,
 	isFollowing: false,
+	locallyUnfollowedFriendIds: [],
 	potentialFriends: [],
 	potentialFriendsTotalCount: 0,
 };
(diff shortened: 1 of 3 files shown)


apps/frontend/src/pages/friends/friends.tsx
```diff
@@ -79,6 +79,12 @@ const Friends: React.FC = () => {
 						page: potentialFriendsPagination.page,
 					}),
 				);
+				void dispatch(
+					actions.getFollowings({
+						count: PaginationValue.DEFAULT_COUNT,
+						page: followingsPagination.page,
+					}),
+				);
```
GEOFARL, line 87: Haven't found a better way with current implementation to update a stale followings data. We do not handle it client side as we do not want to remove friend cards after following or unfollowing them

what1s1ove: you need t refetch friends on the each page after opening all the time. and you need to fetch friends for the opened page only

what1s1ove (changes requested):

apps/frontend/src/modules/friends/slices/friends.slice.ts
```diff
@@ -108,6 +110,24 @@ const { actions, name, reducer } = createSlice({
...
+			state.locallyUnfollowedFriendIds.push(action.payload);
+		},
+		clearFollowings(state) {
+			state.followings = state.followings.filter((following) => {
+				return !state.locallyUnfollowedFriendIds.includes(following.id);
+			});
+		},
+		removeLocallyUnfollowedFriendId(state, action: PayloadAction<number>) {
+			state.locallyUnfollowedFriendIds =
+				state.locallyUnfollowedFriendIds.filter((id) => id !== action.payload);
```
ShadowOfTheSpace, line 129: B27 here

ShadowOfTheSpace (changes requested):

apps/frontend/src/modules/friends/slices/friends.slice.ts
Oleksandra-Masiuk: @ShadowOfTheSpace 
image

ShadowOfTheSpace: @what1s1ove said on one of the daily calls that we will not use that exception for our project right now, so even one-line arrow functions need to have curly braces

Sytnikovakatya (approved):

apps/frontend/src/modules/friends/slices/friends.slice.ts
what1s1ove: it's okay for only the one line function

what1s1ove (approved):

PR #491, opened 2024-03-12
fix(frontend): fix the log out button is shown in two rows tm-480

Vfivew:


apps/frontend/src/pages/user/user.tsx
```diff
@@ -104,7 +104,7 @@ const User: React.FC = () => {
 	}, [dispatch, userId, page]);
 
 	const hasUser = Boolean(profileUser);
-	const hasCommonCourses = commonCourses.length > EMPTY_ARRAY_LENGTH;
+	const hasCommonCourses = commonCourses.length > EMPTY_LENGTH;
```
Vfivew, line 107: The linter was complaining here, so the variable needed to be changed. Everything will work correctly!

fshabanov (approved):

PR #497, opened 2024-03-13
feat(frontend): add transitions during changing the state of the element tm-492

Vfivew:
transition
transition2


fshabanov:
@Vfivew please, add similar transitions (animations) when we are opening popups and other similar parts of the app

what1s1ove (approved):

PR #507, opened 2024-03-13
fix(frontend): Email field on the Auth page has "text" type tm-481

GEOFARL:


what1s1ove (changes requested):
this is expected that we have type=text
use inputmode to change the keyboard

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -33,6 +34,7 @@ const Input = <T extends FieldValues>({
 	errors,
 	hasVisuallyHiddenLabel,
 	iconName,
+	inputMode,
```
fshabanov, line 37:
```suggestion
	inputMode = "text",
```

fshabanov (changes requested):

fshabanov (approved):

PR #509, opened 2024-03-13
fix(frontend): fix position of loader and add loader to the button tm-488

MappariChan:
https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/assets/126669822/e2b3a42f-1d64-4d63-adcf-5ce49106f761

https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/assets/126669822/c21a3857-18c5-4986-a057-2d6f68a7902a

https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/assets/126669822/805d27e0-cafe-4d67-847f-07d138cd943b

https://github.com/BinaryStudioAcademy/bsa-winter-2023-2024-trackmates/assets/126669822/f206d425-6b6f-4845-a8b2-33519e1cf70e


apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -48,6 +51,11 @@ const Button: React.FC<Properties> = ({
 		hasVisuallyHiddenLabel && "visually-hidden",
 	);
 
+	const buttonContentStyles = getValidClassNames(
+		styles["content"],
+		isLoading && styles["content-hidden"],
```
fshabanov, line 56: Can we use `visually-hidden` here?

fshabanov (changes requested):

apps/frontend/src/libs/components/button/button.tsx
MappariChan: No, we can't, because it simply makes it invisible but it still has the physical size, visually-hidden class has `display: none` which has a little bit another logic

fshabanov (approved):

what1s1ove (approved):

PR #532, opened 2024-03-14
feat(frontend): feat add transitions for modal tm-492

Vfivew:
modal


GEOFARL (approved):

apps/frontend/src/libs/components/portal/portal.tsx
```diff
@@ -1,39 +1,11 @@
...
-import styles from "./styles.module.css";
-
 type Properties = {
 	children: React.ReactNode;
 };
 
-const NO_OVERFLOW_CLASS = styles["no-overflow"] as string;
-
 const Portal: React.FC<Properties> = ({ children }: Properties) => {
-	const portalContainer = useMemo(() => {
```
what1s1ove, line 14: we need to use portal for modal

what1s1ove (changes requested):

apps/frontend/src/libs/components/portal/portal.tsx
Vfivew: I dont deleted the Portal.
I just removed unnecessary styles and made it always mounted. 
Otherwise, it won't be possible to achieve animation, because currently the Portal mounts and unmounts from the DOM tree on click.

Ann-Maksay:
Since this is part of ticket 492 (Add transitions during changing the state of the elements), I think it's worth adding transitions to the opening comments as well.

transition-comments

what1s1ove (approved):

PR #539, opened 2024-03-14
fix(frontend): Hide previous state of the data from the store on all pages tm-506

GEOFARL:
Course Progress does not jump no more. It stays at 0 before progress data is loaded


what1s1ove (approved):

apps/frontend/src/pages/course-description/course-description.tsx
```diff
@@ -55,6 +58,8 @@ const CourseDescription: React.FC = () => {
 	const handleClick = useCallback(() => {
 		navigate(BACK_NAVIGATION_STEP);
 		dispatch(courseActions.clearCurrentCourse());
+		dispatch(courseSectionsActions.reset());
```
what1s1ove, line 61: what if the user uses browser navigation? i thing we need to add this action on the page's unmount

what1s1ove (changes requested):

apps/frontend/src/pages/course-description/course-description.tsx
GEOFARL: Yeah, great point. Missed that scenario

ShadowOfTheSpace (approved):

what1s1ove (approved):

PR #547, opened 2024-03-15
fix(frontend): change cursor style over counter badge tm-395

Sytnikovakatya:


Oleksandra-Masiuk (approved):

GEOFARL (approved):

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -55,6 +55,10 @@
 	outline: 3px solid var(--main-white);
 }
 
+.counter:hover {
+	cursor: default;
```
ShadowOfTheSpace, line 59: Is it need to be default or pointer?

Sytnikovakatya: We can't click on this counter.  That's why I use default. Just to change text cursor that we had on arrow

Oleksandra-Masiuk: @Sytnikovakatya please ask QA what behavior is expected in this case

Sytnikovakatya: Yes, I asked QA. It can be default

what1s1ove (approved):

PR #548, opened 2024-03-15
feat(frontend): incoming notification presentation tm-498

Sytnikovakatya:
Incoming notifications are presented within the currently active filter


apps/frontend/src/modules/user-notifications/slices/user-notifications.slice.ts
```diff
@@ -12,12 +12,14 @@ import {
 
 type State = {
 	dataStatus: ValueOf<typeof DataStatus>;
+	notificationType: ValueOf<typeof NotificationFilter> | null;
```
GEOFARL, line 15: Shouldn't it be NotificationFilter.ALL by default instead of null?

Sytnikovakatya: We have on backendside logic if null to use NotificationFilter.ALL

what1s1ove (approved):

PR #562, opened 2024-03-15
feat(frontend): pagination should not be displayed if there is only one page tm-502

Vfivew:
newpag


Sytnikovakatya (approved):

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/check-is-valid-page/check-is-valid-page.helpers.ts
```diff
@@ -0,0 +1,14 @@
+import { PaginationValue } from "~/libs/enums/enums.js";
+
+const checkIsValidPage = (
+	pageFromQuery: number,
+	pagesCount: number,
+): boolean => {
+	const isPageInteger = Number.isInteger(pageFromQuery);
+	const isPageDefault = pageFromQuery > PaginationValue.DEFAULT_PAGE;
+	const isPageExist = pagesCount >= pageFromQuery;
```
YourTeaChan, line 9:
```suggestion
	const isPageExists = pageCount >= pageFromQuery;
```

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/helpers.ts
```diff
@@ -1,2 +1,3 @@
+export { checkIsValidPage } from "./check-is-valid-page/check-is-valid-page.helpers.js";
```
GEOFARL, line 1:
```suggestion
export { checkIsValidPage } from "./check-is-valid-page/check-is-valid-page.helper.js";
```

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -18,17 +23,15 @@ type UsePagination = (options: {
...
 	const [searchParameters, setSearchParameters] = useSearchParams();
 	const pageFromQuery = Number(searchParameters.get(queryName));
-	const isPageInteger = Number.isInteger(pageFromQuery);
-	const isPageValid =
-		isPageInteger && pageFromQuery > PaginationValue.PAGE_NOT_EXISTS;
-	const validPage = isPageValid ? pageFromQuery : PaginationValue.DEFAULT_PAGE;
-	const [page, setPage] = useState<number>(validPage);
-	const [path, setPath] = useState<string>(location.pathname);
+	const [page, setPage] = useState<number>(PaginationValue.DEFAULT_PAGE);
+	const currentPath = path || location.pathname;
```
what1s1ove, line 33: A1

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
```diff
@@ -1,11 +1,16 @@
...
+import {
+	checkIsValidPage,
+	getPagesCut,
+	getPagesRange,
+} from "./libs/helpers/helpers.js";
 
 type UsePagination = (options: {
 	pageSize: number;
 	pagesCutCount: number;
+	path?: string;
```
what1s1ove, line 13:
```suggestion
```
it should not be a part of useHook. this page should work on any page

what1s1ove (changes requested):

apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts
Vfivew: I'll take another look now what I can come up with.
I used the path because we're calling the pagination hook three times in the Friends component. And I have to somehow check my current path against the path I'm making a request to. And only if they match, I should insert the query parameter.

Vfivew: The hook works everywhere.
However, on the friends page, there are tabs, and when we render it, we make requests for three routes (potential friends / followers / followings).
I use an additional parameter (path) to determine whether I need to set a query parameter (page) in the URL, or if this hook was called for a tab that is currently not displayed. If the hook was called only to retrieve information, it will not set the URL query parameter. If the tab is active, then we will set the query parameter.

The issue arises only with the friends page. There, the logic is slightly different from other parts of the application. Should I refactor whole logic in the friends page component to avoid passing "path" in hook?

what1s1ove (approved):

PR #583, opened 2024-03-17
feat(frontend): change buttons in the pagination to links tm-534

Vfivew:


YourTeaChan (approved):

apps/frontend/src/libs/components/pagination/pagination.tsx
```diff
@@ -40,40 +30,40 @@ const Pagination: React.FC<Properties> = ({
 		>
 			<ul className={styles["content"]}>
 				<PaginationItem
-					hasVisuallyHiddenLabel
+					href={`?page=${ONE_ITEM_COUNT}` as ValueOf<typeof AppRoute>}
```
KrasKa2, line 33: Is it valid operation? Do we have `"?page={count}"` in `AppRoute` ?

KrasKa2 (changes requested):

Sytnikovakatya (approved):

apps/frontend/src/libs/components/pagination/pagination.tsx
Vfivew: Yes, of course.
I'm just setting a query parameter in the URL.
it is impossible to set href without type casting
I don't think I need to add a query parameter to AppRoute. Or did you mean something else?

apps/frontend/src/libs/components/link/link.tsx
```diff
@@ -32,7 +35,11 @@ const Link: React.FC<Properties> = ({
 	);
 
 	return (
-		<NavLink className={handleLinkStyles} to={to}>
+		<NavLink
+			className={handleLinkStyles}
+			tabIndex={isDisabled ? TAB_INDEX_DISABLED : undefined}
```
what1s1ove, line 40:
```suggestion
```
you need to render a `span` if the link is in the disabled state

what1s1ove (changes requested):

apps/frontend/src/libs/components/link/link.tsx
Vfivew: I'm rendering span.
I just don’t want the user to be able to interact with this span using the keyboard.
Thats why i did it.
I have another solution.
For example, if `isDisabled ` i can pass href=''. It will resolve my problem. 
But basically I saw the implementation through the tab index.

what1s1ove: if the link is disabled you need to render a span instead
image

what1s1ove (changes requested):
you need to render a span in the `Link` component

what1s1ove (approved):

PR #586, opened 2024-03-17
feat(frontend): pagination should not be displayed if there is only one page tm-502

Vfivew:


fshabanov (approved):

Oleksandra-Masiuk (approved):

apps/frontend/src/libs/components/pagination/pagination.tsx
```diff
@@ -32,6 +32,10 @@ const Pagination: React.FC<Properties> = ({
 		[onPageChange],
 	);
 
+	if (pagesCount === ONE_ITEM_COUNT) {
+		return null;
```
KrasKa2, line 36:
```suggestion
		return <></>;
```

KrasKa2 (changes requested):

apps/frontend/src/libs/components/pagination/pagination.tsx
Vfivew: I don't think it's worth using your syntax.
I have two reasons:
1) We already use returning null in the project.
2) I haven't seen the return of <></> in other projects, but returning null is used.
I'll leave the comment unresolved so that others can participate in the discussion.

Sytnikovakatya (approved):

GEOFARL (approved):

PR #588, opened 2024-03-17
fix(frontend): The "Sign In | TrackMates" page title is shown after trying going back from "Overview" page tm-566

GEOFARL:

### DIFF
--- a/apps/frontend/src/pages/auth/auth.tsx
+++ b/apps/frontend/src/pages/auth/auth.tsx
@@ -1,5 +1,5 @@
 import logo from "~/assets/img/logo.svg";
-import { Image, Navigate } from "~/libs/components/components.js";
+import { Image } from "~/libs/components/components.js";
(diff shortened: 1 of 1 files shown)


Vfivew (approved):

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -52,7 +54,9 @@ const Auth: React.FC = () => {
 	};
 
 	if (hasUser) {
-		return <Navigate to={AppRoute.ROOT} />;
+		navigate(AppRoute.ROOT, { replace: true });
+
+		return;
```
tarmiel, line 59:
```suggestion
		return <Navigate replace to={AppRoute.ROOT} />
```
You can just add replace prop. Why do you need to use useNavigate?

GEOFARL: Yeah, missed it, thanks

what1s1ove (approved):

PR #589, opened 2024-03-17
feat(backend/frontend/shared): update design of activities page tm-555

MappariChan:
image


packages/shared/src/modules/activities/libs/types/activity-counts.type.ts
```diff
@@ -1,5 +1,6 @@
 type ActivityCounts = {
 	commentCount: number;
+	isLikedByUser: null | number;
```
tarmiel, line 3: With is- prefix expected boolean

apps/backend/src/modules/activities/activity.repository.ts
```diff
@@ -22,6 +22,16 @@ class ActivityRepository implements Repository<ActivityEntity> {
 		this.activityModel = activityModel;
 	}
 
+	private checkIsLikedByUserQuery(
```
tarmiel, line 25: Such name supposes to return boolean

tarmiel (changes requested):

ShadowOfTheSpace (approved):

Vfivew (approved):

packages/shared/src/modules/activities/libs/types/activity-response-dto.type.ts
```diff
@@ -6,6 +6,7 @@ import { type ActivityPayloadMap } from "./activity-payload-map.type.js";
 type ActivityResponseDto<T extends ValueOf<typeof ActivityType>> = {
 	commentCount: number;
 	id: number;
+	isLikedByUser: boolean | null;
```
what1s1ove, line 9:
```suggestion
	isLikedByUser: boolean;
```
should be `boolean`. add `?? false` in cases where we need to handle it

apps/frontend/src/pages/feed/libs/components/feed-activity/feed-activity.tsx
```diff
@@ -29,40 +34,60 @@ const FeedActivity: React.FC<Properties> = ({ activity }: Properties) => {
 		void dispatch(activitiesActions.likeActivity(activity.id));
 	}, [dispatch, activity.id]);
 
+	const likeButtonClasses = getValidClassNames(
+		styles["tool-button"],
+		(activity.isLikedByUser as boolean) && styles["liked"],
```
what1s1ove, line 39: but it is not a boolean

what1s1ove (changes requested):

PR #595, opened 2024-03-18
fix(frontend): "Chat" buttons on user components are shifted to the right outside the components [Mobile] tm-565

GEOFARL:
Corrected the chats icon in the header, that was affected by changing the size of the svg icon

image


apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -69,6 +69,11 @@
 	fill: currentcolor;
 }
 
+.icon-button :first-child {
+	width: 24px;
+	height: 24px;
+}
+
```
ShadowOfTheSpace, line 76: Can you change size of icon in friend component? As far as I know, you made that icon smaller in friend component by changing svg file. So can you revert that here and make that icon smaller in friend component where you need it?

what1s1ove: request my once it's done

what1s1ove (changes requested):

what1s1ove (approved):

PR #599, opened 2024-03-18
feat(frontend/backend): add possibility to delete own comments tm-489

tarmiel:


apps/backend/src/libs/hooks/check-is-user-creator/check-is-user-creator.hook.ts
```diff
@@ -0,0 +1,36 @@
+import { type FastifyRequest } from "fastify";
+
+import { ExceptionMessage, HTTPCode } from "~/libs/enums/enums.js";
+import { type Service } from "~/libs/types/types.js";
+import { PermissionError } from "~/modules/permissions/permissions.js";
+
+type CheckIsUserCreatorParameters<T> = {
+	getDtoIdFromRequest: (request: FastifyRequest) => number;
+	getUserIdFromDto: (dto: T) => number;
```
what1s1ove, line 9:
```suggestion
```
provide only paths. not cbs

what1s1ove (changes requested):

apps/backend/src/modules/notifications/notification.service.ts
```diff
@@ -87,10 +87,7 @@ class NotificationService implements Service {
 		);
 
 		if (!notification) {
-			throw new NotificationError({
-				message: ExceptionMessage.NOTIFICATION_NOT_FOUND,
-				status: HTTPCode.NOT_FOUND,
-			});
```
ShadowOfTheSpace, line 93: Why did you remove exception here?

apps/backend/src/libs/hooks/check-is-user-creator/check-is-user-creator.hook.ts
what1s1ove: > What if we have nesting more than one level in dto? Should I handle parsing pathInDto: "author.id" into dto[author][id]

yes, you should handle it

> And what if we need to get the dto id not from params but from the query or body?

pass it as an argument of hook

apps/backend/src/libs/hooks/check-is-user-creator/check-is-user-creator.hook.ts
```diff
@@ -0,0 +1,36 @@
...
+import { type Service } from "~/libs/types/types.js";
+import { PermissionError } from "~/modules/permissions/libs/exceptions/exceptions.js";
+
+type CheckIsUserCreatorParameters<T> = {
+	getDtoIdFromRequest: (request: FastifyRequest) => number;
+	getUserIdFromDto: (dto: T) => number;
+	service: Service<T>;
+};
+
+const checkIsUserCreator =
```
what1s1ove, line 13: it is not only for user. make it more generic

apps/frontend/src/modules/activities/slices/activities.ts
```diff
@@ -1,5 +1,6 @@
 import {
 	createComment,
+	deleteComment,
```
what1s1ove, line 3: it's strange that we have comments here. move all things that are related to comments to `comments` module

what1s1ove (changes requested):

Sytnikovakatya (approved):

packages/shared/src/libs/helpers/get-value-by-path/get-value-by-path.helper.ts
```diff
@@ -0,0 +1,30 @@
...
+	? Left extends keyof TData
+		?
+				| Extract<TData[Left], undefined>
+				| GetFieldType<Exclude<TData[Left], undefined>, Right>
+		: undefined
+	: TPath extends keyof TData
+		? TData[TPath]
+		: undefined;
+
+function getValueByPath<TData, TPath extends string>(
```
what1s1ove, line 11: arrow function

apps/backend/src/libs/hooks/check-is-equal/check-is-equal.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+	}: CheckIsEqualParameters<DtoType, RequestType>) =>
+	async (request: FastifyRequest): Promise<void> => {
+		const toCompareId = getValueByPath(
+			request as RequestType,
+			pathInRequestToCompareId,
+		) as number;
+		const dtoId = getValueByPath(
+			request as RequestType,
+			pathInRequestToDtoId,
+		) as number;
```
what1s1ove, line 30:
```suggestion
		const dtoId = getValueByPath<number>(
			request as RequestType,
			pathInRequestToDtoId,
		); // number or null
		
		if (!dtoId) {
		  // error
		}
```

apps/backend/src/libs/hooks/check-is-equal/check-is-equal.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const checkIsEqual =
+	<DtoType, RequestType extends FastifyRequest>({
+		pathInDtoToCompareId,
+		pathInRequestToCompareId,
+		pathInRequestToDtoId,
+		service,
+	}: CheckIsEqualParameters<DtoType, RequestType>) =>
+	async (request: FastifyRequest): Promise<void> => {
+		const toCompareId = getValueByPath(
```
what1s1ove, line 23:
```suggestion
		const compareId = getValueByPath(
```

apps/backend/src/libs/hooks/check-is-equal/check-is-equal.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+			request as RequestType,
+			pathInRequestToCompareId,
+		) as number;
+		const dtoId = getValueByPath(
+			request as RequestType,
+			pathInRequestToDtoId,
+		) as number;
+
+		const dto = await service.find(dtoId);
+		const dtoIdToCompare = getValueByPath(dto, pathInDtoToCompareId);
```
what1s1ove, line 33:
```suggestion
		const compareDtoId = getValueByPath(dto, pathInDtoToCompareId);
```

apps/backend/src/libs/hooks/check-is-equal/check-is-equal.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+import { PermissionError } from "~/modules/permissions/libs/exceptions/exceptions.js";
+
+type CheckIsEqualParameters<DtoType, RequestType extends FastifyRequest> = {
+	pathInDtoToCompareId: ExtractPath<DtoType>;
+	pathInRequestToCompareId: ExtractPath<RequestType>;
+	pathInRequestToDtoId: ExtractPath<RequestType>;
+	service: Service<DtoType>;
+};
+
+const checkIsEqual =
```
what1s1ove, line 15: do not like the name. should be something like this checkIsEqual<WHO> ?

apps/backend/src/libs/hooks/check-is-equal/check-is-equal.hook.ts
what1s1ove: same check for compareId

what1s1ove (changes requested):

apps/backend/src/libs/hooks/check-is-equal/check-is-equal.hook.ts
tarmiel: I do not like it either, but I'm not completely sure what is the exact responsibility of this hook

what1s1ove (approved):

PR #609, opened 2024-03-18
fix(frontend): The Back button does not navigate to the previous page on the Friend's profile page tm-535

GEOFARL:


what1s1ove (changes requested):
Fix CI and then I will merge

what1s1ove (approved):

PR #614, opened 2024-03-19
fix(frontend): Add titles for the Friend profile and Not found pages tm-601

GEOFARL:
image
image


fshabanov (changes requested):
Let's display friend name if we are in a friend profile screen

Sytnikovakatya (approved):

what1s1ove (approved):

PR #619, opened 2024-03-19
fix(frontend/backend): UAM (User Access Management) - Groups tm-230

YourTeaChan:


apps/frontend/src/pages/management/libs/components/groups-tab/groups-tab.tsx
```diff
@@ -233,12 +233,14 @@ const GroupsTab: React.FC = () => {
 							});
 
 						return (
-							<li className={styles["modal-item"]} key={permission.id}>
+							<li
```
what1s1ove, line 236: move this modal to a separate component

what1s1ove (changes requested):

apps/frontend/src/pages/management/libs/components/groups-tab/groups-tab.tsx
```diff
@@ -233,12 +233,14 @@ const GroupsTab: React.FC = () => {
...
+								className={styles["modal-item"]}
+								key={`${currentGroup?.id}${permission.id}`}
```
what1s1ove, line 238: it can not be null, is not it?

what1s1ove (changes requested):

apps/frontend/src/pages/management/libs/components/groups-tab/groups-tab.tsx
YourTeaChan: `currentGroup` can't be `null`, when modal is open, but by default and after modal closing `currentGroup` is `null`. So, I need this `?` here

apps/frontend/src/pages/management/libs/components/groups-tab/libs/components/edit-group-modal/edit-group-modal.tsx
```diff
@@ -0,0 +1,105 @@
...
+	type GroupResponseDto,
+	actions as groupsActions,
+} from "~/modules/groups/groups.js";
+import { type PermissionResponseDto } from "~/modules/permissions/permissions.js";
+import { EditCheckbox } from "~/pages/management/libs/components/edit-checkbox/edit-checkbox.jsx";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	group: GroupResponseDto | null;
```
what1s1ove, line 15:
```suggestion
	group: GroupResponseDto;
```
it can not be null since this popup is opened only when you have a `group` in state, is not it?

apps/frontend/src/pages/management/libs/components/users-tab/libs/components/edit-user-modal/edit-user-modal.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { EditCheckbox } from "~/pages/management/libs/components/edit-checkbox/edit-checkbox.jsx";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	groups: GroupResponseDto[];
+	isOpen: boolean;
+	onClose: () => void;
+	title?: string | undefined;
+	user: UserAuthResponseDto | null;
```
what1s1ove, line 17: same

what1s1ove (changes requested):

apps/frontend/src/pages/management/libs/components/groups-tab/libs/components/edit-group-modal/edit-group-modal.tsx
YourTeaChan: @what1s1ove 
After adding animation for popups, the popup renders once when the page loads, not every time when the popup opens. To not pass a null value to custom popups, I need to use conditional rendering. But it will disable popup animation.
What should I leave in this case? Popup animation or passing not null values?

apps/frontend/src/pages/management/libs/components/users-tab/libs/components/edit-user-modal/edit-user-modal.tsx
```diff
@@ -0,0 +1,90 @@
...
+									return userGroup.id === group.id;
+								}),
+							);
+
+							return (
+								<li className={styles["checkbox-item"]} key={group.id}>
+									<EditCheckbox
+										isChecked={isChecked}
+										itemId={group.id}
+										key={group.id}
```
fshabanov, line 66:
```suggestion
```

fshabanov (changes requested):

apps/frontend/src/pages/management/libs/components/groups-tab/libs/components/edit-group-modal/edit-group-modal.tsx
what1s1ove: both. try to find a way to fix it

YourTeaChan: @what1s1ove 
I tried different ways to keep an animation and not to pass null values, but as a result, I received errors of reading null values or non-animated popup. I asked for help from my teammate, but the result was the same.
One idea that works - it's to pass a some default user or group to popup if the current user is null. But, I think, it's a bad solution, isn't it?
I pushed this solution, so I will wait for your answer.

apps/frontend/src/pages/management/libs/components/groups-tab/groups-tab.tsx
```diff
@@ -213,41 +196,14 @@ const GroupsTab: React.FC = () => {
 					</Table>
 				</div>
 			</div>
-			<EditModal
+			<EditGroupModal
+				group={currentGroup ?? DEFAULT_GROUP}
```
KrasKa2, line 200:
```suggestion
     {currentGroup &&
     			<EditGroupModal
		      		group={currentGroup}
```
maybe, it is better?

KrasKa2 (changes requested):

apps/frontend/src/pages/management/libs/components/groups-tab/groups-tab.tsx
YourTeaChan: Conditional rendering removes popup animation, but I need to keep it.

apps/frontend/src/pages/management/libs/components/groups-tab/libs/constants/default-group.constant.ts
```diff
@@ -0,0 +1,10 @@
+import { type GroupResponseDto } from "~/modules/groups/groups.js";
+
+const DEFAULT_GROUP = {
```
what1s1ove, line 3:
```suggestion
```
it's bad to have just mocked data

it's better to remove the animation from popup then

or render it differently. Ex `group ** <Popup group={group} />`

what1s1ove (changes requested):

apps/frontend/src/pages/management/libs/components/groups-tab/libs/constants/default-group.constant.ts
YourTeaChan: Okay, I will use conditional rendering

what1s1ove (approved):

PR #666, opened 2024-03-21
feat(backend/frontend/shared): Add pagination to the Activities page tm-646

GEOFARL:


apps/backend/src/modules/activities/activity.repository.ts
```diff
@@ -242,6 +247,78 @@ class ActivityRepository implements Repository<ActivityEntity> {
 		});
 	}
 
+	public async findAllWithPagination({
```
what1s1ove, line 250:
```suggestion
	public async findAllWith({
```
should be with params. it's by default with pagination

apps/frontend/src/pages/feed/libs/components/feed-activity-list/feed-activity-list.tsx
```diff
@@ -11,20 +12,61 @@ import styles from "./styles.module.css";
...
+		[handleIncreasePage, isLoadingMore],
+	);
+
 	return (
 		<div className={styles["feed"]}>
 			{activities.map((activity) => (
 				<FeedActivity activity={activity} key={activity.id} userId={user.id} />
 			))}
+
+			{activities.length !== totalCount && (
```
what1s1ove, line 65: add constant

apps/frontend/src/pages/feed/libs/components/feed-activity-list/feed-activity-list.tsx
```diff
@@ -11,20 +12,61 @@ import styles from "./styles.module.css";
...
+}: Properties) => {
 	const { user } = useAppSelector(({ auth }) => {
 		return {
 			user: auth.user as UserAuthResponseDto,
 		};
 	});
 
+	const observer = useRef<IntersectionObserver | null>(null);
+
+	const loaderElementReference = useCallback(
```
what1s1ove, line 34: add a custom hook for it

apps/frontend/src/modules/activities/activities-api.ts
```diff
@@ -22,17 +23,27 @@ class ActivitiesApi extends BaseHTTPApi {
 		super({ baseUrl, http, path: APIPath.ACTIVITIES, storage });
 	}
 
-	public async getActivities(): Promise<ActivityGetAllResponseDto> {
+	public async getActivities(
```
what1s1ove, line 26:
```suggestion
	public async getAll(
```

what1s1ove (changes requested):

apps/backend/src/modules/activities/activity.repository.ts
```diff
@@ -242,6 +247,78 @@ class ActivityRepository implements Repository<ActivityEntity> {
 		});
 	}
 
+	public async findAllWithParams({
```
what1s1ove, line 250:
```suggestion
	public async findAll({
```

apps/frontend/src/libs/hooks/use-infinite-scroll/use-infinite-scroll.hook.ts
```diff
@@ -0,0 +1,40 @@
+import { useCallback, useRef } from "react";
+
+type Properties = {
+	isLoading: boolean;
+	loadMore: () => void;
```
what1s1ove, line 5:
```suggestion
	onLoadMore: () => void;
```

what1s1ove (changes requested):

apps/backend/src/modules/activities/activity.repository.ts
GEOFARL: image
image
image

what1s1ove: implement them. it should be this route, right

what1s1ove (changes requested):

apps/backend/src/modules/activities/activity.repository.ts
GEOFARL: @what1s1ove, but how should I implement them? FindAll method expects to return only the array of entities, not the response in the pagination format with items and total. No repository that supports pagination implements it in the findAll method

GEOFARL: @what1s1ove, but what to do with `total` in that case? I need it either

what1s1ove: you can add it to the return type in the repository. what is the problem?

what1s1ove (approved):

PR #670, opened 2024-03-21
fix(frontend): Courses on the "Overview" page disappear and reappear, when a new course is being added on the "Add course" pop-up window tm-610

GEOFARL:


apps/frontend/src/modules/user-courses/libs/helpers/check-whether-add-course.helper.ts
```diff
@@ -0,0 +1,14 @@
+import { PaginationValue } from "~/libs/enums/enums.js";
+
+const checkWhetherAddCourse = (
```
what1s1ove, line 3:
```suggestion
const checkIsWhetherAddCourse = (
```
filename

what1s1ove (changes requested):
I guess just we need to refetch courses once adding popup is closed. it will be the best way to handle it. let's do this way. remove all inecessary changes

what1s1ove (approved):

PR #678, opened 2024-03-21
feat(frontend): update pop up window for downloading pwa tm-672

Vfivew:
pwascrreen


GEOFARL (approved):

apps/frontend/public/assets/screen-desk-fourth.jpg
fshabanov, line 1: Isn't it the same with `screen-desk-first.jpg`?

fshabanov (changes requested):

apps/frontend/public/assets/screen-desk-fourth.jpg
Vfivew: This screenshot has another dimensions. If we not added this one, the screenshot will not be available on devices with small screens.

what1s1ove (approved):

PR #686, opened 2024-03-22
fix(frontend): The "Log Out" button can't be reached on the sidebar in landscape mode [Mobile] tm-567

GEOFARL:


apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -78,6 +78,17 @@ const Sidebar: React.FC<Properties> = ({ menuItems, user }: Properties) => {
...
+								className={getValidClassNames(
+									styles["burger-button-in-flow"],
+									styles[isOpen ? "open" : "close"],
+								)}
+								hasVisuallyHiddenLabel
+								iconName="burger"
+								label="burger-button"
+								onClick={handleToggleSidebar}
+								style="secondary"
+							/>
```
what1s1ove, line 91:
```suggestion
```
you can not duplicate the markup

what1s1ove (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar.tsx
GEOFARL: @what1s1ove 
I cannot achieve the same working sidebar with all the desired behaviours with only one burger button, not sure whether I'll be able to do that till the code freeze, but the fixes are really important

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -58,21 +61,25 @@ const Sidebar: React.FC<Properties> = ({ menuItems, user }: Properties) => {
 		[user],
 	);
 
+	const windowWidth = useWindowWidth();
```
what1s1ove, line 64: bad solution. we need to handle it only with css. without any js

what1s1ove (changes requested):

apps/frontend/src/libs/components/sidebar/styles.module.css
```diff
@@ -294,12 +321,12 @@
...
-	.title-container {
-		max-height: 46px;
+	.content-container {
+		gap: 14px;
 	}
 
-	.sidebar .content-wrapper {
-		padding-right: 12px;
+	.content-wrapper {
+		overflow: hidden scroll;
```
tarmiel, line 329:
```suggestion
		overflow: hidden auto;
```

tarmiel (changes requested):

PR #694, opened 2024-03-22
fix(frontend): course stable layout and default image tm-668

Sytnikovakatya:
- Introduced a default image feature for courses, which will be displayed if the course image fails to load 

Снимок экрана (37)

- Prevented layout shifts

**Before**

Снимок экрана (46)

**After**

Снимок экрана (50)


apps/frontend/src/libs/components/course/libs/component/course-card/course-card.tsx
```diff
@@ -16,15 +18,30 @@ type Properties = {
 const CourseCard: React.FC<Properties> = ({ course }: Properties) => {
 	const { image, title, vendor } = course;
 
+	const [isImageLoaded, setIsImageLoaded] = useState<boolean>(true);
+
+	const [imageSource, setImageSource] = useState<string>(
+		image || defaultCourseImage,
+	);
```
what1s1ove, line 25: we should incapsulate this logic by adding a `default` image prop to the Image component

what1s1ove (changes requested):

what1s1ove (approved):

